PR #1, opened 2022-04-08
Get Protofiles From `serum-api`

kdonthi:


Makefile
```diff
@@ -0,0 +1,15 @@
+IMAGE_NAME=bloxroute/bdn-protobuf:v3.19.3-rp-gateway
+LOCAL_IMAGE_NAME=bx-proto-gen
+
+.PHONY: proto
+proto: proto-build-api proto-build-gw
+
+.PHONY: proto-build-gw
+proto-build-gw:
```
aspin: i **think** this is unnecessary? unless it also generates http client stubs?

aspin (approved):

PR #2, opened 2022-04-08
BX-4015 Work on Orderbook SDK

kdonthi:
* Creating GetOrderbook and GetOrderbookStream for clients to use in library


bxserum/provider/ws.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				response, err := recvWSResponse[T](conn)
```
kdonthi: I can put this inside a go routine and get the result from a channel, but thought it was unnecessary

bxserum/provider/ws.go
```diff
@@ -0,0 +1,112 @@
...
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbook", "params": {"market":"%s"}}`, market)
+	return unaryWSRequest[pb.GetOrderbookResponse](w.conn, request)
+}
+
+func (w *WSClient) GetOrderbookStream(ctx context.Context, market string, orderbookChan chan *pb.GetOrderbookStreamResponse) error {
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbookStream", "params": {"market":"%s"}}`, market)
+	return unaryWSStream[pb.GetOrderbookStreamResponse](ctx, w.conn, request, orderbookChan)
+}
+
+func unaryWSRequest[T any](conn *websocket.Conn, request string) (*T, error) {
```
kdonthi: We can create two type sets of the stream and non-stream responses to restrict what types can be passed in

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/gorilla/websocket"
+	"github.com/sirupsen/logrus"
+	"github.com/sourcegraph/jsonrpc2"
+)
+
+func UnaryWSRequest[T any](conn *websocket.Conn, request string) (*T, error) {
```
kdonthi: Create type sets for allowed types?

bxserum/provider/ws.go
```diff
@@ -0,0 +1,44 @@
...
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbook", "params": {"market":"%s"}}`, market)
+	return helpers.UnaryWSRequest[pb.GetOrderbookResponse](w.conn, request)
+}
+
+func (w *WSClient) GetOrderbookStream(ctx context.Context, market string, orderbookChan chan *pb.GetOrderbookStreamResponse) error {
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbookStream", "params": {"market":"%s"}}`, market)
+	return helpers.UnaryWSStream[pb.GetOrderbookStreamResponse](ctx, w.conn, request, orderbookChan)
+}
+
+func (w *WSClient) CloseConn() error {
```
aspin: `Close()` is enough

cmd/wsclient/main.go
```diff
@@ -0,0 +1,44 @@
...
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
+)
+
+func main() {
+	callWebsocket()
+}
+
+func callWebsocket() {
+	w, err := provider.NewWSClient("ws://174.129.154.164:1810/ws")
```
aspin: It'd probably be a good idea for the default client to have our address hardcoded, e.g.

```
provider.NewWSClient() // connects to production
provider.NewWSClientTestnet() // doesn't exist yet
provider.NewWSClientWithEndpoint("") // if you want to manually specify
```

cmd/wsclient/main.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		log.Fatal("dial:", err)
+	}
+	defer w.CloseConn()
+
+	// One time request
+	orderBook, err := w.GetOrderbook("ETH/USDT")
```
aspin: minor: orderbook as one word

bxserum/provider/ws.go
```diff
@@ -0,0 +1,44 @@
...
+	return &WSClient{addr: addr, conn: conn}, nil
+}
+
+func (w *WSClient) GetOrderbook(market string) (*pb.GetOrderbookResponse, error) {
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbook", "params": {"market":"%s"}}`, market)
+	return helpers.UnaryWSRequest[pb.GetOrderbookResponse](w.conn, request)
+}
+
+func (w *WSClient) GetOrderbookStream(ctx context.Context, market string, orderbookChan chan *pb.GetOrderbookStreamResponse) error {
+	request := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "GetOrderbookStream", "params": {"market":"%s"}}`, market)
```
aspin: it'd probably be a good idea to write some methods around this to construct the `jsonrpc2` object instead of writing the string directly

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,76 @@
...
+	err := sendWSRequest(conn, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return recvWSResult[T](conn)
```
aspin: One problem you're going to experience right now is that you're going to have issues interweaving unary requests and streams. I think this is fine as a v0.1, but you may find that your requests get intercepted between each other. To solve this you're going to need to create a single go routine that dispatches all read messages to either a waiting request handler or subscribed stream. I think let's not solve this right away but create a ticket and add a big fat TODO on this file for that. I have the same issue in Python right now: https://bloxroute.atlassian.net/browse/BX-4123

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,76 @@
...
+func UnaryWSStream[T any](ctx context.Context, conn *websocket.Conn, request string, responseChan chan *T) error {
+	err := sendWSRequest(conn, request)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
```
aspin: this is nice :+1:

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, fmt.Errorf("error reading WS response - %v", err)
+	}
+
+	// extract the result
+	var resp jsonrpc2.Response
+	err = resp.UnmarshalJSON(msg)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling JSON response - %v", err)
+	}
+	resultBytes, err := resp.Result.MarshalJSON()
```
aspin: I'd like to avoid this if possible. What type does `jsonrpc2.Response` deserialize to? We're probably going to want it to be `json.RawMessage`, which will skip any deserialization, and then you can call `Unmarshal` on it again.

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,76 @@
+package helpers
```
aspin: I think I would want the contents of this file being in `ws.go`, or a closely related file like `provider/ws_conn.go`. I'm not sure yet, but I don't want a `helpers` module to grow too large

aspin (changes requested):
Generally looks good! I'd suggest fixing the `json.RawMessage` thing now and making a follow up ticket for message separation

bxserum/helpers/ws.go
kdonthi: Sounds good and I added a TODO.

I think that having unique request IDs for each request is important for that, so I went ahead and created a function to get/increment the request ID for each request using a lock.

bxserum/helpers/ws.go
kdonthi: If I understand your question, the `jsonrpc2.Response` is a struct that has fields for the ID, Result, and Error. The issue with replacing it with `json.RawMessage` is that I will not be able to separate the Result field as easily from it. When I tried replacing the struct, I got empty values in the response structs.

bxserum/helpers/ws.go
kdonthi: Do you mean I should keep it in one of those places or I shouldn't? I can also rename the folder to `requests` if that helps?

bxserum/helpers/ws.go
kdonthi: :)

bxserum/helpers/ws.go
kdonthi: One thing I tried to do was to create an in function struct like this:
```
type response struct {
    Response T
}
```
but I got the error that `type declarations inside generic functions are not currently supported`

aspin: Yeah sorry what I want is only the `Result` field to be `json.RawMessage`.

So something like

```go
type response struct {
    ID string
    Result json.RawMessage
    Error error
}

// later

json.Unmarshal(response.Result, &impl)
```

which delay deserializing the `Result` field specifically

bxserum/helpers/ws.go
aspin: Fixed the comment –– sorry I didn't type that out well. Basically, the logic in here isn't really related to each other, and as it grows will end up wanting to be three different packages, I think (`http`, `ws` and `grpc`). We can hold off on refactoring that until later though.

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/sourcegraph/jsonrpc2"
+	"sync"
+)
+
+// TODO Handle sending responses to their correct locations
+
+var requestID uint64 = 1
+var requestIDLock sync.Mutex
+
+func GetRequestID() uint64 {
```
aspin: `RequestID()` (no Get)

bxserum/helpers/ws.go
```diff
@@ -0,0 +1,93 @@
...
+
+	requestIDLock.Lock()
+	val = requestID
+	requestID++
+	requestIDLock.Unlock()
+
+	return val
+}
+
+func UnaryWSRequest[T any](conn *websocket.Conn, request []byte) (*T, error) {
```
aspin: Do you like `Unary` in the name of this function? I honestly don't really know why GRPC calls these functions `unaryUnary` and `unaryStream`. I think it might be nicer if we separated all these into `ws`, `http`, and `grpc`, then just called this `Request` and `Stream`. So the usage would be `ws.Request` and `ws.Stream`. 

Alternatively, since we might need some struct eventually to track each of the different ws streams, we might want to define an interface for each of the types with `Request` and `Stream` methods respectively. wdyt?

aspin: in the second case I'd suggest the module be `connections`?

aspin (commented):
`provider/ws.go` looks :+1:

bxserum/helpers/ws.go
kdonthi: Got it, I believe the jsonrpc2.Response has a `json.RawMessage` for its Result.
```
type Response struct {
    ID     ID               `json:"id"`
    Result *json.RawMessage `json:"result,omitempty"`
    Error  *Error           `json:"error,omitempty"`
    Meta   *json.RawMessage `json:"meta,omitempty"`
}
```

But I think I will create a new struct like you suggested; dereferencing a pointer to a RawMessage makes me uneasy.

bxserum/helpers/ws.go
kdonthi: I wanted to use Unary to be consistent with the Python SDK, but I think removing it will be better for naming. I also like the name `connections` for the package name.

An interface and different structs would be nice, but currently methods cannot have type parameters. Maybe, we can keep track of the streams inside the client structs (i.e. WSClient)?

bxserum/connections/ws.go
```diff
@@ -0,0 +1,92 @@
...
+
+	requestIDLock.Lock()
+	val = requestID
+	requestID++
+	requestIDLock.Unlock()
+
+	return val
+}
+
+func WSRequest[T any](conn *websocket.Conn, request []byte) (*T, error) {
```
aspin, line 18: The main concern I expect to have about all of this eventually is that the state is all global. I really like having the generic `recWSResult[T]` function, but I don't want the final state of this to have a global channel for processing all websocket messages. It should be possible for the user to create multiple websocket clients and have them all be dependent on each other. But I think fine for this PR for now.

bxserum/connections/ws.go
```diff
@@ -0,0 +1,92 @@
...
+// TODO Handle sending responses to their correct locations
+type response struct {
+	Result json.RawMessage
+	Error  jsonrpc2.Error
+}
+
+var requestID uint64 = 1
+var requestIDLock sync.Mutex
+
+func RequestID() uint64 {
```
aspin: thought about this some more and I don't think this should be global either. Probably should just be an attribute on the websocket provider (or if you want it to be its own stryct with only a `Next()` function that would be a good candidate for a `utils` type package). Should probably also simplify this function just a little bit and use `defer`

```go
requestIDLock.Lock()
defer requestIDLock.Unlock()

val := requestID
requestID++
return val
```

aspin (changes requested):
just a couple more

bxserum/connections/ws.go
kdonthi: Did you mean dependent or independent?

aspin: independent** my bad

utils/request_id.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+func NewRequestID() RequestID {
+	return RequestID{
+		id:   1,
+		lock: &sync.Mutex{},
+	}
+}
+
+func (r RequestID) Next() uint64 {
```
aspin: this needs a pointer receiver I think

bxserum/connections/ws.go
aspin: remove this code now?

bxserum/connections/ws.go
kdonthi: I think that since we have a different connection for each client, that should be fine?

aspin: well where do you envision the global channel for handling all subscription messages to live? The code as is it currently is doesn't allocated this global channel for each connection. You'd have to pass it in on each function call and that's a bit clunky to pass in both a global channel and a channel for subscription events (for example)

aspin (approved):

PR #4, opened 2022-04-22
Tests and Updates for GetMarkets, GetOrderbook, and GetOrderbookStream 

kdonthi:
* Created tests for WS, GRPC, and HTTP clients
* Added `limit` as a parameter to GetOrderbook methods
* Handle streaming errors in WS/GRPC by handling error from first request


bxserum/connections/http.go
```diff
@@ -3,41 +3,75 @@ package connections
...
+func HTTPGetWithClient[T protoreflect.ProtoMessage](url string, client *http.Client, val T) error {
 	httpResp, err := client.Get(url)
 	if err != nil {
-		return nil, err
+		return err
+	}
+	if httpResp.StatusCode != 200 {
+		return httpUnmarshalError(httpResp)
+	}
+
```
kdonthi, line 40: remove newl

bxserum/connections/http.go
```diff
@@ -3,41 +3,75 @@ package connections
...
 }
 
 // HTTP response for GET request
-func HTTPGetWithClient[T any](url string, client *http.Client) (*T, error) {
+func HTTPGetWithClient[T protoreflect.ProtoMessage](url string, client *http.Client, val T) error {
 	httpResp, err := client.Get(url)
 	if err != nil {
-		return nil, err
+		return err
+	}
```
kdonthi, line 35: add newl

bxserum/connections/http.go
```diff
@@ -3,41 +3,75 @@ package connections
...
+	if httpResp.StatusCode != 200 {
+		return httpUnmarshalError(httpResp)
```
kdonthi, line 38: is this always going to be an error?

bxserum/connections/http.go
```diff
@@ -3,41 +3,75 @@ package connections
...
-		return nil, err
+		return err
 	}
 
-	var resp T
-	if err := json.Unmarshal(b, &resp); err != nil {
-		return nil, err
+	if err := protojson.Unmarshal(b, val); err != nil {
+		return err
 	}
```
kdonthi, line 79: newl

bxserum/connections/ws.go
```diff
@@ -69,7 +70,11 @@ func recvWSResult[T any](conn *websocket.Conn) (*T, error) {
 		return nil, fmt.Errorf("error unmarshalling JSON response - %v", err)
 	}
 	if resp.Error.Data != nil {
-		return nil, fmt.Errorf("error in JSON response - %s", resp.Error.Message)
+		m, err := json.Marshal(resp.Error.Data)
+		if err != nil {
+			return nil, err
+		}
```
kdonthi, line 81: newl

bxserum/provider/grpc.go
```diff
@@ -39,19 +39,23 @@ func NewGRPCClientWithEndpoint(endpoint string) (*GRPCClient, error) {
 	}, nil
 }
 
-func (g *GRPCClient) GetOrderbook(ctx context.Context, market string) (*pb.GetOrderbookResponse, error) {
-	return g.apiClient.GetOrderbook(ctx, &pb.GetOrderBookRequest{Market: market})
+func (g *GRPCClient) GetOrderbook(ctx context.Context, market string, limit uint32) (*pb.GetOrderbookResponse, error) {
```
kdonthi, line 42: add comment for how limit == 0 gets all the bids/asks

bxserum/provider/http.go
```diff
@@ -27,15 +27,33 @@ func NewHTTPTestnet() *HTTPClient {
 	panic("implement me")
 }
 
-// Connects to custom Serum API (set client to nil to use default client)
+// Connects to custom Serum API
```
kdonthi:
```suggestion
// Connects to custom Serum API (set client to nil to use default client)
```

bxserum/provider/http.go
```diff
@@ -27,15 +27,33 @@ func NewHTTPTestnet() *HTTPClient {
...
 		client = &http.Client{Timeout: time.Second * 7}
 	}
 	return &HTTPClient{baseURL: endpoint, httpClient: client}
 }
 
-func (h *HTTPClient) GetOrderbook(market string) (*pb.GetOrderbookResponse, error) {
+func (h *HTTPClient) GetOrderbook(market string, limit uint32) (*pb.GetOrderbookResponse, error) {
 	url := h.baseURL + fmt.Sprintf("/api/v1/market/orderbooks/%s", market)
-	return connections.HTTPGetWithClient[pb.GetOrderbookResponse](url, h.httpClient)
+	if limit != 0 {
```
kdonthi: what if you add the limit either way?

bxserum/provider_test/integration/grpc_test.go
```diff
@@ -0,0 +1,56 @@
...
+			return markets
+		},
+	)
+}
+
+func TestGRPCClient_Streams(t *testing.T) {
+	g, err := provider.NewGRPCClient()
+	require.Nil(t, err)
+
+	TestGetOrderbookStream(
```
kdonthi: should there be an error func for this too?

kdonthi: no errors for stream?

bxserum/provider_test/integration/public.go
```diff
@@ -0,0 +1,117 @@
...
+	pb "github.com/bloXroute-Labs/serum-api/proto"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+const publicRequestTimeout = 5 * time.Second
+
+// if limit == -1: assume no limit
```
kdonthi, line 15: Does limit == -1 mean no limit? or 0?...

bxserum/provider_test/integration/public.go
```diff
@@ -0,0 +1,117 @@
...
+	ctx, cancel = context.WithTimeout(context.Background(), publicRequestTimeout)
+	defer cancel()
+	orderbook = getOrderbookFn(ctx, "SOL-USDC", 0)
+
+	assertSOLUSDCOrderbook(t, "SOL-USDC", orderbook)
+
+	// use limit
+	ctx, cancel = context.WithTimeout(context.Background(), publicRequestTimeout)
+	defer cancel()
+	orderbook = getOrderbookFn(ctx, "SOL-USDC", 2)
```
kdonthi, line 39: should we use diff market names?

bxserum/provider_test/integration/streaming.go
```diff
@@ -0,0 +1,46 @@
...
+	for i := 0; i < streamExpectEntries; i++ {
+		orderbook := ReadChan[*pb.GetOrderbookStreamResponse](t, orderbookCh, streamExpectTimeout)
+		require.NotNil(t, orderbook)
+
+		assertSOLUSDCOrderbook(t, "SOL/USDC", orderbook.Orderbook)
+	}
+	cancel()
+}
+
+func ReadChan[T any](t *testing.T, c chan T, timeout time.Duration) T {
```
kdonthi: move this into bxassert

bxserum/provider_test/integration/streaming.go
```diff
@@ -0,0 +1,46 @@
...
+	orderbookCh := make(chan *pb.GetOrderbookStreamResponse)
+	go connectFn(ctx, "SOLUSDC", 0, orderbookCh)
+
+	for i := 0; i < streamExpectEntries; i++ {
+		orderbook := ReadChan[*pb.GetOrderbookStreamResponse](t, orderbookCh, streamExpectTimeout)
+		require.NotNil(t, orderbook)
+
+		assertSOLUSDCOrderbook(t, "SOL/USDC", orderbook.Orderbook)
+	}
+	cancel()
```
kdonthi: try to see if you can include errors?

bxserum/provider_test/integration/ws_test.go
```diff
@@ -0,0 +1,54 @@
...
+	)
+
+	testGetMarkets(
+		t,
+		func(ctx context.Context) *pb.GetMarketsResponse {
+			markets, err := w.GetMarkets()
+			require.Nil(t, err)
+
+			return markets
+		})
```
kdonthi: more last paren down?

examples/grpcclient/main.go
```diff
@@ -21,16 +21,19 @@ func callGRPC() {
 	}
 
 	// Unary response
-	orderbook, err := g.GetOrderbook(context.Background(), "ETH-USDT")
+	ctx, cancel := context.WithCancel(context.Background())
```
kdonthi: we don't really need context?

go.mod
```diff
@@ -4,18 +4,21 @@ go 1.18
...
 	github.com/sourcegraph/jsonrpc2 v0.1.0
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd
 	google.golang.org/genproto v0.0.0-20220317150908-0efb43f6373e
 	google.golang.org/grpc v1.45.0
 	google.golang.org/protobuf v1.27.1
 )
 
 require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
kdonthi, line 17: do you need these?

bxserum/provider/http.go
```diff
@@ -35,7 +35,23 @@ func NewHTTPClientWithEndpoint(endpoint string, client *http.Client) *HTTPClient
...
+		return nil, err
+	}
+
+	return orderbook, nil
+}
+
+func (h *HTTPClient) GetMarkets() (*pb.GetMarketsResponse, error) {
+	url := h.baseURL + fmt.Sprintf("/api/v1/market/markets")
+	markets := new(pb.GetMarketsResponse)
+	if err := connections.HTTPGetWithClient[*pb.GetMarketsResponse](url, h.httpClient, markets); err != nil {
```
kdonthi, line 52: variable ordering?

bxserum/provider_test/integration/public.go
```diff
@@ -0,0 +1,117 @@
+package integration
```
kdonthi, line 1: same as public.go in serum-api

bxserum/provider_test/integration/stream.go
```diff
@@ -0,0 +1,45 @@
+package integration
+
+import (
+	"context"
+	"github.com/bloXroute-Labs/serum-api/bxserum/provider_test/bxassert"
```
kdonthi: very similar to stream.go in serum-api except for the bxassert import and cleanup of a few unneeded lines at the bottom of the test

examples/wsclient/main.go
```diff
@@ -31,7 +30,7 @@ func callWS() {
 
 	fmt.Println()
 
-	orderbook, err = w.GetOrderbook("SOLUSDT")
+	orderbook, err = w.GetOrderbook("SOLUSDT", 2)
```
kdonthi, line 34: make sure these are working

bxserum/connections/ws.go
```diff
@@ -62,14 +94,18 @@ func recvWSResult[T any](conn *websocket.Conn) (*T, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading WS response - %v", err)
 	}
-
-	// extract the result
+	// extract the HTTP Response Result
```
kdonthi: newl

bxserum/provider/ws.go
```diff
@@ -29,44 +27,43 @@ func NewWSClientTestnet() (*WSClient, error) {
...
+	connectionManager := connections.NewConnectionManager(addr)
 
 	return &WSClient{
-		addr:      addr,
-		conn:      conn,
-		requestID: utils.NewRequestID(),
-	}, nil
+		addr:              addr,
+		connectionManager: &connectionManager,
+		requestID:         utils.NewRequestID(),
```
kdonthi: Should I get rid of the request ID? I think Websockets might be the only place where handling connections is needed?

bxserum/connections/ws.go
```diff
@@ -15,37 +16,68 @@ type response struct {
...
+
+	err = sendWSRequest(conn, request)
 	if err != nil {
 		return err
 	}
 
-	go func() {
+	response, err := recvWSResult[T](conn)
+	if err != nil {
+		log.Errorf("error in ws stream %v", err)
```
kdonthi: handle conn before the function ends

bxserum/connections/ws.go
kdonthi: HTTP?

bxserum/connections/grpc.go
```diff
@@ -1 +1,43 @@
...
+			} else {
+				responseChan <- response
+			}
+		}
+	}(response, stream, input)
+
+	return nil
+}
+
+func recvGRPCResponse[T any](stream grpc.ClientStream, input string) (*T, error) {
```
aspin: `recvGRPC`

bxserum/connections/grpc.go
```diff
@@ -1 +1,43 @@
 package connections
+
+import (
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"io"
+)
+
+func GRPCStream[T any](stream grpc.ClientStream, input string, responseChan chan *T) error {
```
aspin, line 9: I think you'll want a `ctx` argument for this, so you can cancel the stream by canceling the ctx

bxserum/connections/http.go
```diff
@@ -3,41 +3,78 @@ package connections
...
+	if err := httpUnmarshal[T](httpResp, val); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func httpUnmarshalError(httpResp *http.Response) error {
+	if httpResp == nil {
+		return fmt.Errorf("HTTP response is nil")
```
aspin: minor: these should be `errors.New`. Might also be a good idea to define some error constants for these

bxserum/connections/http.go
```diff
@@ -3,41 +3,78 @@ package connections
...
+	}
+
+	var httpError HTTPError
+	body, err := ioutil.ReadAll(httpResp.Body)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling response to HTTPError")
```
aspin: this is probably ok to just return the `err` directly

bxserum/connections/http.go
```diff
@@ -3,41 +3,78 @@ package connections
...
+
+	return nil
+}
+
+func httpUnmarshalError(httpResp *http.Response) error {
+	if httpResp == nil {
+		return fmt.Errorf("HTTP response is nil")
+	}
+
+	var httpError HTTPError
```
aspin: move this var declaration down to line 56

bxserum/connections/ws_connection_manager.go
```diff
@@ -0,0 +1,51 @@
...
+func NewConnectionManager(address string) ConnectionManager {
+	return ConnectionManager{
+		connectionMap: make(map[int]*websocket.Conn),
+		address:       address,
+		id:            0,
+	}
+}
+
+func (c *ConnectionManager) Next() (*websocket.Conn, int, error) {
+	conn, _, err := websocket.DefaultDialer.Dial(c.address, nil)
```
aspin: general best not to use the `DefaultDialer`. Allocate a new one for the manager so we don't have to mess with the global settings

bxserum/provider/http.go
```diff
@@ -35,7 +35,23 @@ func NewHTTPClientWithEndpoint(endpoint string, client *http.Client) *HTTPClient
 	return &HTTPClient{baseURL: endpoint, httpClient: client}
 }
 
-func (h *HTTPClient) GetOrderbook(market string) (*pb.GetOrderbookResponse, error) {
-	url := h.baseURL + fmt.Sprintf("/api/v1/market/orderbooks/%s", market)
-	return connections.HTTPGetWithClient[pb.GetOrderbookResponse](url, h.httpClient)
+// Set limit to 0 to get all bids/asks
+func (h *HTTPClient) GetOrderbook(market string, limit uint32) (*pb.GetOrderbookResponse, error) {
+	url := h.baseURL + fmt.Sprintf("/api/v1/market/orderbooks/%s?limit=%v", market, limit)
```
aspin: best not to intermix string concat and `fmt.Sprintf`. Just do `fmt.Sprintf("%v/api/...")`

bxserum/provider/http.go
```diff
@@ -35,7 +35,23 @@ func NewHTTPClientWithEndpoint(endpoint string, client *http.Client) *HTTPClient
...
+	orderbook := new(pb.GetOrderbookResponse)
+	if err := connections.HTTPGetWithClient[*pb.GetOrderbookResponse](url, h.httpClient, orderbook); err != nil {
+		return nil, err
+	}
+
+	return orderbook, nil
+}
+
+func (h *HTTPClient) GetMarkets() (*pb.GetMarketsResponse, error) {
+	url := h.baseURL + fmt.Sprintf("/api/v1/market/markets")
```
aspin: same here

bxserum/connections/ws_connection_manager.go
```diff
@@ -0,0 +1,51 @@
+package connections
+
+import (
+	"fmt"
+	"github.com/gorilla/websocket"
+)
+
+type ConnectionManager struct {
```
aspin: `connections.ConnectionManager` stutters a bit. I'd suggest `connections.WsManager`

bxserum/provider/ws.go
```diff
@@ -29,44 +27,43 @@ func NewWSClientTestnet() (*WSClient, error) {
...
+func NewWSClientWithEndpoint(addr string) *WSClient {
+	connectionManager := connections.NewConnectionManager(addr)
 
 	return &WSClient{
-		addr:      addr,
-		conn:      conn,
-		requestID: utils.NewRequestID(),
-	}, nil
+		addr:              addr,
+		connectionManager: &connectionManager,
```
aspin: `NewConnectionManager` should probably already return a pointer

bxserum/connections/ws.go
```diff
@@ -15,37 +16,70 @@ type response struct {
...
 		return nil, err
 	}
 
 	return recvWSResult[T](conn)
 }
 
-func WSStream[T any](ctx context.Context, conn *websocket.Conn, request []byte, responseChan chan *T) error {
-	err := sendWSRequest(conn, request)
+func WSStream[T any](ctx context.Context, connectionManager *ConnectionManager, request []byte, responseChan chan *T) error {
+	conn, id, err := connectionManager.Next()
```
aspin: This isn't quite right. Are you trying to create a new websocket connection each time we start streaming? That's going to be quite a bit of overhead we don't want. One stream should be capable of handling multiple subscriptions. Do we have the appropriate setup in the websockets server for this yet? (e.g. assigning each subscription a subscription ID so we can do the mapping?) Something like this: https://github.com/bloXroute-Labs/bxcommon-private/blob/develop/src/bxcommon/rpc/provider/response_queue.py

bxserum/connections/ws.go
```diff
@@ -63,13 +97,18 @@ func recvWSResult[T any](conn *websocket.Conn) (*T, error) {
 		return nil, fmt.Errorf("error reading WS response - %v", err)
```
aspin: ^ suggest shortening to `sendWS` and `recvWS`

bxserum/connections/ws.go
```diff
@@ -63,13 +97,18 @@ func recvWSResult[T any](conn *websocket.Conn) (*T, error) {
...
 	}
 
-	// extract the result
+	// Extract the WS Response Result
```
aspin: comment got refactored? looks a little weird

aspin (commented):
a couple of smaller things but main problem is the opening of multiple websocket connections per provider, which shouldn't be necessary

bxserum/connections/grpc.go
kdonthi: The reason I did not think I needed to was because the context is already taken when creating a stream. When the context is canceled, RecvMsg should return an error and we will return out of the for loop.

bxserum/provider/http.go
kdonthi: Is it for consistency or are there errors that can happen because of this?

aspin: consistency. `fmt.Sprintf` is much more readable. Apparently this is a bit slower, but we don't need to prematurely optimize. Don't really expect this to be a bottleneck anyway.

bxserum/connections/grpc.go
aspin: gotcha, sounds good :+1:

bxserum/connections/grpc.go
```diff
@@ -1 +1,43 @@
...
+		return err
+	}
+
+	go func(response *T, stream grpc.ClientStream, input string) {
+		responseChan <- response
+
+		for {
+			response, err = recvGRPC[T](stream, input)
+			if err != nil {
+				logrus.Errorf(err.Error())
```
aspin: It'd usually be best not log this, as the person that uses this library may not be using `logrus` and might not want the logs. I haven't really thought about the best way to handle this. Can you create a ticket in Serum 3 epic for thinking about handling logs in each of the client libraries?

bxserum/connections/ws.go
```diff
@@ -15,49 +16,57 @@ type response struct {
 	Error  jsonrpc2.Error
 }
 
-func WSRequest[T any](conn *websocket.Conn, request []byte) (*T, error) {
-	err := sendWSRequest(conn, request)
+func WSResponse[T any](conn *websocket.Conn, request []byte) (*T, error) {
```
aspin: This should be `WSRequest`. You're sending and receiving a response –– this is a request.

bxserum/connections/ws.go
```diff
@@ -15,49 +16,57 @@ type response struct {
...
-	err := sendWSRequest(conn, request)
+	err := sendWS(conn, request)
 	if err != nil {
 		return err
 	}
 
-	go func() {
+	response, err := recvWS[T](conn)
+	if err != nil {
+		log.Errorf("error in ws stream %v", err)
```
aspin: same comments –– for here since the error can be returned directly though I'd just do that

bxserum/connections/ws.go
```diff
@@ -15,49 +16,57 @@ type response struct {
...
 			}
 		}
-	}()
+	}(response, responseChan, conn)
 
 	return nil
 }
 
-func sendWSRequest(conn *websocket.Conn, request []byte) error {
+func sendWS(conn *websocket.Conn, request []byte) error {
```
aspin, line 59: nits: a couple of general notes around punctuation in error messages: technically hyphens (`-`) are not the appropriate for this. Hyphens are supposed to be used only for joining words, like "self-restraint." The right punctuation is an em dash –– we should just use colons instead, which is much more conventional

bxserum/provider_test/integration/stream.go
```diff
@@ -0,0 +1,46 @@
...
+
+		assertSOLUSDCOrderbook(t, "SOL/USDC", orderbook.Orderbook)
+	}
+	cancel()
+
+	// unknown market
+	ctx, cancel = context.WithCancel(context.Background())
+	defer cancel()
+
+	errMessage := connectFnErr(ctx, "market-doesnt-exist", 0) // #2. cleaned up code here
```
aspin, line 44: fix comment?

aspin (approved):
- some naming / punctuation comments
- remove log messages from client library

generally looks :+1:

bxserum/connections/ws.go
kdonthi: I agree - I was just thinking that WSStream sends a request as well, so I thought using `Response` would separate them better

kdonthi: Should I use `WSStreamRequest` instead of `WSStream`?

aspin: Request is still better –– it describes the action that is being performed. It's true that streaming involves request, but you could also say that streaming involves responses. Request usually refers to 1 request/response pair. (If anything streaming is 1 request and n responses, so it's more response than this one)

bxserum/connections/grpc.go
kdonthi: Sure, I can do that.

bxserum/provider_test/integration/stream.go
kdonthi: I wanted to keep those comments to show the difference between the stream.go file in serum-api. I thought it could help if someone is updating this file

bxserum/connections/ws.go
aspin: Nah, `WSStream` is better. Short names are best when possible.

bxserum/provider_test/integration/stream.go
kdonthi: I think it might be nice to have a central location that both repositories could access the tests at some point

aspin: gotcha, I see. It's not very obvious to me here though.

aspin: Yes, I agree with the second part of that later. But let's figure that out at a later date.

bxserum/connections/grpc.go
aspin: Not sure if you edited, but I agree with the concern about this being in a goroutine and the error not being exposed somewhere. Add a note in the ticket about it.

PR #6, opened 2022-05-09
[BX-4142] Add mapping for ws requests.

Jeeyb:


bxserum/provider/ws.go
```diff
@@ -2,71 +2,139 @@ package provider
...
+	"sync"
 )
 
+type requests struct {
+	sync.Mutex
+	dataReceivers map[uint64]chan json.RawMessage
+}
+
+func (r *requests) add(id uint64, receiver chan json.RawMessage) {
+	r.Mutex.Lock()
```
aspin: since your embedding the mutex in `r`, you can just do `r.Lock()`

bxserum/provider/ws.go
```diff
@@ -2,71 +2,139 @@ package provider
...
+
+type subscriptions struct {
+	*requests
+	subscriptions       map[string]chan json.RawMessage
+	subscriptionsReader chan connections.SubscriptionResponse
+}
+
+func (s *subscriptions) read() {
+	for response := range s.subscriptionsReader {
+		log.Println(response)
```
aspin: delete this line?

bxserum/provider/ws.go
```diff
@@ -2,71 +2,139 @@ package provider
...
 	if err != nil {
 		return nil, err
 	}
-	return connections.WSRequest[pb.GetOrderbookResponse](w.conn, request)
+
+	responseChan := make(chan json.RawMessage)
+	w.requests.add(requestID, responseChan)
+	defer w.requests.delete(requestID)
+
+	return connections.WSRequest[pb.GetOrderbookResponse](w.conn, request, responseChan)
```
aspin: This sequence of calls looks like it can be refactored into a single function?

aspin: I'm not sure it makes sense to keep `connections/ws.go` in general –– it might be better to move most of that logic into this struct. It's just a bit weird to call a function `WSRequest`, then give that function the response channel to which the calling code intends to write the response to. From the signature of the `WSRequest` `responseChan` looks like it should be getting the response written to it –– something like

```go
connections.WSRequest[pb.GetOrderbookResponse](w.conn, request, responseCh)
<-responseCh
```

aspin: I would also consider writing a wrapper struct in `connections/ws.go` and leaving `provider/ws.go` completely unaware of the mechanics behind request IDs / subscription IDs. Something like

```go
type WsRPC interface {
    Request(methodName string, message json.RawMessage) (json.RawMessage, error)
    Stream(message json.RawMessage, responses chan json.RawMessage) error // provider/ws.go doesn't really need to know about subscription IDs
}

func (w *wsRPC) read() { ... }

func (w *wsRPC) Request(methodName string, message []byte) json.RawMessage {
    id := w.nextID()
    rpcMessage := jsonrpc2.Request { ID: id, Method: methodName, Params: message }
    w.conn.Send(rpcMessage)
    
    response, err := w.wait(id)
    return response.Result, err
}

// provider/ws.go
func (w *WSClient) GetOrderbook(market string, limit uint32) (*pb.GetOrderbookResponse, error) {
    params := map[string]interface{}{"market": market, "limit": limit})

    result, err := w.wsRPC.Request("GetOrderbook", params)
    if err != nil {
        return nil, err
    }

    var response pb.GetOrderbookResponse
     json.Unmarshal(...)
}
```

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
+			dataReceivers: map[uint64]chan json.RawMessage{},
+		},
+		streamSubscriptions: subscriptions{
+			requests:            &requests{dataReceivers: map[uint64]chan json.RawMessage{}},
+			subscriptions:       map[string]chan json.RawMessage{},
+			subscriptionsReader: make(chan subscriptionResponse, 1000),
+		},
+	}
+
+	go wsRPC.read()
```
Jeeyb, line 79: Probably we need send message to customer if something went wrong, mb add something like error chan ? @aspin WDYT?

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
-	Error  jsonrpc2.Error
+type requests struct {
+	sync.Mutex
+	dataReceivers map[uint64]chan json.RawMessage
 }
 
-func WSRequest[T any](conn *websocket.Conn, request []byte) (*T, error) {
-	err := sendWS(conn, request)
+func (r *requests) add(id uint64, receiver chan json.RawMessage) {
+	r.Mutex.Lock()
```
aspin, line 19: `r.Lock()` `defer r.UnLock()`. Though I don't think you really want to make `r.Mutex` visible outside of this struct? IMO should be a private property `m` or something

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
+	"sync"
 )
 
-// TODO Handle sending responses to their correct locations
-type response struct {
-	Result json.RawMessage
-	Error  jsonrpc2.Error
+type requests struct {
+	sync.Mutex
+	dataReceivers map[uint64]chan json.RawMessage
```
aspin, line 15: naming suggestion: `dataReceiver` sounds like you can receive data on these channels. Suggest something like `requestListeners` or `requestWaiters`

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
+	SubscriptionID string          `json:"subscriptionID"`
+	Params         json.RawMessage `json:"params"`
+}
+
+type WSRPC struct {
+	requestID utils.RequestID
+	conn      *websocket.Conn
+
+	streamSubscriptions subscriptions
+	nonStreamRequests   requests
```
aspin, line 62: ok to just call this `requests`: request already implies not a stream

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
+	subscriptionsReader chan subscriptionResponse
+}
+
+func (s *subscriptions) read() {
+	for response := range s.subscriptionsReader {
+		_, ok := s.subscriptions[response.Result]
+		if ok {
+			continue
+		}
+		s.subscriptions[response.Result] = s.dataReceivers[response.ID]
```
aspin, line 42: I think this is a possible race condition? What if `delete` is called at the same time as this line? I think it should be possible to code this all out without any mutexs if you're just using a single loop to do all the reading/writing

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
 				if err != nil {
-					break
+					return err
 				}
 
-				responseChan <- response
+				return errors.New(string(m))
 			}
+
+			return errors.New("failed to check ")
```
aspin, line 161: fill this out? would be a malformed error –– could probably just serialize the whole thing an include it

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
 		}
-	}(response, responseChan, conn)
 
-	return nil
-}
+		if resp.ID.Str == "null" {
```
aspin, line 164: This isn't a good guarantee –– the initial subscription response with the subscription ID can have a response ID. It's the subsequent responses that don't have a request ID.

bxserum/connections/ws.go
```diff
@@ -1,92 +1,205 @@
...
+		// new subscription for feed
+		subscriptionResponse := subscriptionResponse{}
+		err = json.Unmarshal(msg, &subscriptionResponse)
+		if err == nil {
+			subscriptionResponse.ID = resp.ID.Num
+			w.streamSubscriptions.subscriptionsReader <- subscriptionResponse
+			continue
+		}
+
+		responseReader, ok := w.nonStreamRequests.dataReceivers[resp.ID.Num]
```
aspin, line 198: I think you should probably do this part before trying to unmarshall the subscription format

bxserum/connections/ws.go
aspin: Yeah, I think so. But a dedicated error channel would be clumsy to work with, wouldn't it? Should probably make all the `chan json.RawMessage` types instead be a struct like this 

```
type RPCResponse {
    result *json.RawMessage
    err error
}
```

Then in the request side you could check each of  these fields to see which one is set?

bxserum/connections/ws.go
aspin: It's a little bit confusing that the receiving channel is moved from `s.request` to `s.subscriptions`. Add a comment to clarify what's happening.

aspin (changes requested):

bxserum/connections/ws.go
Jeeyb: null means that we already have subscription ID

aspin:
continuing in other ticket

PR #7, opened 2022-05-10
BX-4212 Transaction Signing

kdonthi:
- Created function to sign transaction messages
- Created sample code to test transaction signing


bxserum/transaction/signing.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"encoding/base64"
+	"fmt"
+	bin "github.com/gagliardetto/binary"
+	"github.com/gagliardetto/solana-go"
+	"os"
+)
+
+// SignTxMessage uses the environment variable for `PRIVATE_KEY` to sign the transaction
```
kdonthi:
```suggestion
// SignTxMessage uses the environment variable for `PRIVATE_KEY` to sign the transaction message
```

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/base64"
+	"fmt"
+	bin "github.com/gagliardetto/binary"
+	"github.com/gagliardetto/solana-go"
+	"os"
+)
+
+// SignTxMessage uses the environment variable for `PRIVATE_KEY` to sign the transaction message
+func SignTxMessage(txMessage string) (string, error) {
+	pKeyStr := os.Getenv("PRIVATE_KEY")
```
aspin: minor: probably slightly better is `pkeyStr, ok := os.LookupEnv`

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,50 @@
...
+	if pKeyStr == "" {
+		return "", fmt.Errorf("env variable `PRIVATE_KEY` not set")
+	}
+	pKey, err := solana.PrivateKeyFromBase58(pKeyStr)
+	if err != nil {
+		return "", err
+	}
+
+	var m solana.Message
+	txBytes, err := base64.StdEncoding.DecodeString(txMessage)
```
aspin: As mentioned in Slack –– should decode a full `solana.Transaction` here

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,50 @@
...
+	err = signTx(solanaTx, pKey)
+	if err != nil {
+		return "", err
+	}
+
+	return solanaTx.ToBase64()
+}
+
+func signTx(tx *solana.Transaction, privateKey solana.PrivateKey) error {
+	sigs, err := tx.Sign(func(key solana.PublicKey) *solana.PrivateKey {
```
aspin: `Sign` overwrites private keys –– need to append (pending findings on if signature order matters)

bxserum/transaction/signing.go
aspin: Transaction will also come with recent block hashes set when returned from API

bxserum/transaction/signing.go
kdonthi: I think Sign just appends to the signatures (but the way I had written `signTx` previously did overwrite the signatures; I changed it now).

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,54 @@
...
+		if key.Equals(privateKey.PublicKey()) {
+			return &privateKey
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
aspin: can be simplified to `return err`

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func signTx(tx *solana.Transaction, privateKey solana.PrivateKey) error {
+	signaturesRequired := int(tx.Message.Header.NumRequiredSignatures)
+	signaturesPresent := len(tx.Signatures)
+	if signaturesPresent != signaturesRequired-1 {
+		return fmt.Errorf("transaction requires %v signatures and has %v signatures, should need exactly one more signature", signaturesRequired, signaturesPresent)
+	}
+
+	_, err := tx.Sign(func(key solana.PublicKey) *solana.PrivateKey {
```
aspin: I think you're already aware of this, but will need to go through the signatures directly and replace the zeroed signatures `tx.Signatures[0].IsZero` with the private key since the ordering matters.

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+	"github.com/gagliardetto/solana-go"
+	solanarpc "github.com/gagliardetto/solana-go/rpc"
+	"os"
+)
+
+// SignTx uses the environment variable for `PRIVATE_KEY` to sign the transaction
+func SignTx(unsignedTxBase64 string) (string, error) {
```
aspin, line 12: Probably should expose another function `SignTxWithPrivateKey` that accepts a private key argument in case the user wants to do their own thing with the loading the key.

bxserum/transaction/signing.go
```diff
@@ -1,50 +1,84 @@
...
 }
 
-func signTx(tx *solana.Transaction, privateKey solana.PrivateKey) error {
-	sigs, err := tx.Sign(func(key solana.PublicKey) *solana.PrivateKey {
-		if key.Equals(privateKey.PublicKey()) {
-			return &privateKey
-		}
-		return nil
-	})
+func signTx(solanaTx *solana.Transaction, privateKey solana.PrivateKey) error {
```
kdonthi, line 46: naming?

examples/transaction/main.go
```diff
@@ -0,0 +1,121 @@
...
+	recipient := solana.MustPublicKeyFromBase58(recipientAddress)
+
+	tx, err := solana.NewTransaction([]solana.Instruction{
+		system.NewTransferInstruction(1, publicKey, recipient).Build(),
+	}, recentBlockHash.Value.Blockhash)
+	if err != nil {
+		return nil, err
+	}
+
+	tx.Signatures = append(tx.Signatures, solana.Signature{}) //adding a zero signature
```
kdonthi:
```suggestion
	tx.Signatures = append(tx.Signatures, solana.Signature{}) // adding a zero signature
```

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,84 @@
...
+
+	signedMessageContent, err := privateKey.Sign(messageContent)
+	if err != nil {
+		return fmt.Errorf("unable to sign message: %v", err)
+	}
+
+	found := false
+	for i, sig := range tx.Signatures {
+		if sig.IsZero() {
+			tx.Signatures[i] = signedMessageContent
```
kdonthi: Should we check that there is only one zero signature?

aspin: yeah, I think that's a good idea

bxserum/transaction/signing.go
```diff
@@ -0,0 +1,80 @@
...
+	signaturesRequired := int(solanaTx.Message.Header.NumRequiredSignatures)
+	signaturesPresent := len(solanaTx.Signatures)
+	if signaturesPresent != signaturesRequired {
+		return fmt.Errorf("transaction requires %v signatures and has %v signatures", signaturesRequired, signaturesPresent)
+	}
+
+	return replaceZeroSignature(solanaTx, privateKey)
+}
+
+func replaceZeroSignature(tx *solana.Transaction, privateKey solana.PrivateKey) error {
```
aspin, line 56: probably good to add a comment to this function indicating that the API returns a message with a placeholder for the SDK to complete signing

aspin (approved):

PR #11, opened 2022-05-25
BX 4154 implement post cancel order endpoint

rafaelvanoni:


bxserum/provider/grpc.go
```diff
@@ -113,21 +114,58 @@ func (g *GRPCClient) PostSubmit(ctx context.Context, txBase64 string) (*pb.PostS
...
 	if err != nil {
 		return "", err
 	}
 
-	txBase64, err := transaction.SignTxWithPrivateKey(order.Transaction, g.privateKey)
+	return g.signAndSubmit(ctx, order.Transaction)
+}
+
+// PostCancelOrder builds a Serum cancel order, signs and submits it to the network.
+func (g *GRPCClient) PostCancelOrder(ctx context.Context, orderID string, side pb.Side, owner, market, openOrders string) (string, error) {
```
aspin: something here that I think would be good idea –– separate these two functions out into `SubmitCancelOrder` and `PostCancelOrder`, like is done with `SubmitOrder`. Same elsewhere –– it'd be good to have `Submit` be the name of all functions that actually submit the transaction

aspin (changes requested):

aspin (approved):

PR #13, opened 2022-06-01
BX-4318 Endpoints cleanup

DimaCantemir:
BX-4318 Endpoints cleanup


rafaelvanoni (approved):
Changes to the Cancel* bits lgtm

PR #14, opened 2022-06-09
BX-4346 add option to skip preflight checks

rafaelvanoni:


aspin (approved):

bxserum/provider/http.go
```diff
@@ -253,7 +253,7 @@ func (h *HTTPClient) SubmitCancelByClientOrderID(
 		return "", err
 	}
 
-	return h.signAndSubmit(order.Transaction)
+	return h.signAndSubmit(order.Transaction, true)
```
kdonthi: Do we want to give the option to enable preflight checks to the user for any transaction?

aspin: Actually yeah, I think we do. I'd be fine with either making it an RPCOpt to set on the whole provider or individually specified per endpoint

rafaelvanoni: From what I've seen the Cancel* endpoints always fail if these are enabled, so specifying by endpoint would probably be better than by provider. Will make these changes.

rafaelvanoni: Ok, pushed those changes, please take a look

aspin (approved):

PR #18, opened 2022-06-18
Pre Release Fixes to GO SDK

kdonthi:
* Fixing WS Stream code to not panic
* Adding functions to clients' sample codes
   * reusing clients when possible in sample code

### DIFF
--- a/bxserum/connections/ws.go
+++ b/bxserum/connections/ws.go
@@ -20,13 +20,12 @@ func WSRequest[T proto.Message](conn *websocket.Conn, request []byte, response T
 	return recvWS[T](conn, response)
 }
 
-func WSStream[T proto.Message](ctx context.Context, conn *websocket.Conn, request []byte, responseChan chan T) error {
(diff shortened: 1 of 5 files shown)


bxserum/connections/ws.go
```diff
@@ -20,13 +20,12 @@ func WSRequest[T proto.Message](conn *websocket.Conn, request []byte, response T
 	return recvWS[T](conn, response)
 }
 
-func WSStream[T proto.Message](ctx context.Context, conn *websocket.Conn, request []byte, responseChan chan T) error {
+func WSStream[T proto.Message](ctx context.Context, conn *websocket.Conn, request []byte, responseChan chan T, response T) error {
```
aspin, line 23: This change doesn't make sense –– why would the user provide a response variable for a stream, when they're going to receive all messages on the channel? What compile error is this preventing?

aspin (changes requested):
examples/ changes seem fine. I don't think the change in `connections/ws.go` should go in.

aspin (approved):
I see the problem. Fine for now, but let's create a ticket to try and resolve this

PR #22, opened 2022-06-30
BX-4239 Order Status Feed

kdonthi:
* Created WS and GRPC functions for order status streaming
* Added sample code combining the existing order placement functions


aspin (approved):

bxserum/provider_test/integration/stream.go
```diff
@@ -42,7 +41,23 @@ func testGetOrderbookStream(
...
+		errMessage := connectFnErr(ctx, "market-doesnt-exist", 0)
 		assert.Equal(t, "provided market name/address was not found", errMessage)
 	}
 }
+
+func testGetOrderStatusStream(t *testing.T, connectFnErr func(ctx context.Context, market string, ownerAddress string) string) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	ctx, cancel = context.WithCancel(context.Background())
```
aspin: duplicate line

bxserum/provider_test/integration/ws_test.go
```diff
@@ -94,11 +95,27 @@ func TestWSClient_Requests(t *testing.T) {
 				_, err := w.SubmitOrder(owner, payer, market, side, []pb.OrderType{pb.OrderType_OT_LIMIT}, amount, price, opts)
 				require.NotNil(t, err)
 
-				return err.Error()
+				return strings.Trim(err.Error(), "\"")
```
aspin: don't think this should be necessary -- if you're seeing this it's usually because you cast the error message directly to a `string` without JSON unmarshaling. best to fix this at the source instead of covering it up here

aspin (commented):
changes look good

PR #24, opened 2022-07-11
Bx 4513/add missing endpoints support

DimaCantemir:
Added missed GET Account Balance to HTTP provider


bxserum/provider/http.go
```diff
@@ -111,6 +111,17 @@ func (h *HTTPClient) GetUnsettled(market string, owner string) (*pb.GetUnsettled
 	return result, nil
 }
 
+// GetAccountBalance returns all OpenOrders accounts for a given market with the amounts of unsettled funds
+func (h *HTTPClient) GetAccountBalance(owner string) (*pb.GetAccountBalanceResponse, error) {
+	url := fmt.Sprintf("%s/api/v1/account/balance?ownerAddress=%s", h.baseURL, owner)
+	result := new(pb.GetAccountBalanceResponse)
```
aspin, line 117: I'm actually a little surprised this works in our codebase. Something like this doesn't work for `json.Unmarshal`, since this is a pointer pointing to no allocated data, so there's no data structure to unmarshal the data to, but I suppose `protojson` must do something special w/reflection to get this working.

aspin (approved):

kdonthi (approved):

PR #26, opened 2022-07-13
BX-4111 RPC spec compliance

aspin:


bxserum/connections/ws.go
```diff
@@ -5,84 +5,325 @@ import (
...
+}
+
+func (w *WS) readLoop() {
+	defer w.cancel()
+
+	for {
+		// if lock is held by another processing routine, then wait for release before processing any further messages
+		// typically, messages can be dispatched independently, but in some cases it may be important to finish processing a single message before allowing the socket to read any more (e.g. subscription response before processing potential updates)
+		w.messageM.Lock()
+		w.messageM.Unlock()
```
DimaCantemir, line 70: defer ?

bxserum/connections/ws.go
```diff
@@ -5,84 +5,325 @@ import (
...
+		// try response format first
+		var response jsonrpc2.Response
+		err = json.Unmarshal(msg, &response)
+		if err == nil && (response.Result != nil || response.Error != nil) {
+			w.processRPCResponse(response)
+			continue
+		}
+
+		// if not, try subscription format
+		var update jsonrpc2.Request
```
DimaCantemir, line 87: jsonrpc2.Request ? should it be FeedUpdate ?

DimaCantemir: I guess not, seems like I'm missing something here.

examples/wsclient/main.go
```diff
@@ -36,9 +35,9 @@ func main() {
 	callAccountBalanceWS(w)
 
 	// streaming methods
-	callOrderbookWSStream()
-	callFilteredOrderbookWSStream()
-	callTradesWSStream()
+	callOrderbookWSStream(w)
+	//callFilteredOrderbookWSStream(w)
+	//callTradesWSStream(w)
```
DimaCantemir, line 42: Did you mean to leave the above 2 lines commented?

DimaCantemir (changes requested):

examples/wsclient/main.go
aspin: fixed

bxserum/connections/ws.go
aspin: yeah, so "update" and "request" are the same shape as each other in JSON-RPC. There are only two formats in the JSON-RPC spec, "request" and "response. "response" is only ever in response to a "request", so subscription updates can't be of the "response" format. Instead, it's a "request" without a `requestID`, which indicates that it doesn't expect a "response" in return.

Example:

**Request:**
```
wscat -c ws://127.0.0.1:9000/ws --wait 1000 --execute '{"jsonrpc": "2.0", "id": 1, "method": "subscribe", "params": ["GetOrderbooksStream", {"market": "SOL/USDC", "limit": 10}]}'
```

**Response** (subscribe confirmation):
```
{"id":1,"result":"3c911616-0a8c-4184-a733-138c163f5c03","jsonrpc":"2.0"}
```

**Event** (update "request" with no ID):
```
{"method":"subscribe","params":{"subscription":"3c911616-0a8c-4184-a733-138c163f5c03","result":{"blockHeight":"141322189","orderbook":{"market":"SOL/USDC","marketAddress":"9wFFyRfZBsuAha4YcuxcXLKwMxJR43S7fPfQLusDBzvT","bids":[{"price":32.916,"size":1289.4},{"price":32.93,"size":3.3},{"price":32.943,"size":1507},{"price":32.956,"size":643.9},{"price":32.976,"size":321.7},{"price":32.986,"size":160.8},{"price":33.091,"size":603.2},{"price":33.135,"size":200},{"price":33.137,"size":90.4},{"price":33.138,"size":11.1}],"asks":[{"price":33.165,"size":160.8},{"price":33.175,"size":321.6},{"price":33.195,"size":643.2},{"price":33.2,"size":233.1},{"price":33.217,"size":0.1},{"price":33.218,"size":3},{"price":33.227,"size":0.1},{"price":33.23,"size":90.4},{"price":33.235,"size":1286.5},{"price":33.25,"size":1615.6}]}}},"jsonrpc":"2.0"}
```

bxserum/connections/ws.go
aspin: Not quite. All reading from the websocket is done on a single goroutine, and each message processing happens on a different one, so messages don't necessarily get processed in order. In most cases this is fine since we're just mapping request IDs to responses, but for subscriptions we're going to drop some initial subscription updates if the response isn't processed first (since the map entry won't be setup correctly).

In this implementation, `w.processRPCResponse(response)` may acquire the lock, which will prevent subsequent execution of this loop until it's released. The goroutine that actually handles the messages becomes responsible for indicating when it's done holding the message processing lock. I'll flesh out this comment a bit more.

DimaCantemir (approved):

PR #29, opened 2022-07-22
BX-4348 - Adding MIT License

DimaCantemir:
Should make pkg.go.dev show Readme file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Alexey Palazhchenko
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


aspin (approved):

LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Alexey Palazhchenko
```
aspin: wait this line isn't rigth

aspin (changes requested):

LICENSE
aspin: 2022 bloXroute Labs

DimaCantemir: Fixed

aspin (approved):

PR #34, opened 2022-08-01
code supporting sdn auth header and local deployment

huangzhen1997:
adding auth-header for all request. added local environment


bxserum/connections/http.go
```diff
@@ -32,7 +34,9 @@ func HTTPGet[T protoreflect.ProtoMessage](url string, val T) error {
 }
 
 func HTTPGetWithClient[T protoreflect.ProtoMessage](url string, client *http.Client, val T) error {
-	httpResp, err := client.Get(url)
+	req, err := http.NewRequest("GET", url, nil)
+	req.Header.Set("Authorization", utils.AuthHeader)
```
aspin: shouldn't be hardcoded. I'd include this with `RPCOpts` struct

examples/httpclient/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 }
 
 func callMarketsHTTP() {
-	h := provider.NewHTTPClient()
+	h := provider.NewHTTPLocal()
```
aspin: revert provider code in examples

examples/grpcclient/main.go
```diff
@@ -14,7 +16,7 @@ import (
 )
 
 func main() {
-	g, err := provider.NewGRPCTestnet()
+	g, err := provider.NewGRPCLocal()
```
aspin: revert provider in examples

examples/grpcclient/main.go
```diff
@@ -59,7 +61,10 @@ func main() {
 }
 
 func callMarketsGRPC(g *provider.GRPCClient) {
-	markets, err := g.GetMarkets(context.Background())
+	authHeader := make(map[string]string)
```
aspin: it'd also be a good idea to make all of this part of `RPCOpts`, and include the lines when creating the GRPCProvider

aspin (changes requested):

bxserum/connections/http.go
aspin: bump @huangzhen1997

utils/constant.go
```diff
@@ -0,0 +1,3 @@
+package utils
+
+const AuthHeader = "Your-Auth-Header"
```
aspin: This constant shouldn't exist. I don't think `const` can even be changed by users in Go? When users use this code they're not adding code into this repository. They're importing this library and using the functions we export, so it's not at all easy for them to discover where to set the auth header. You should instead force them to set the auth header in such a way that it's obviously how they can change it –– either by requiring them to include it as a parameter in the function call, or loading it from an environment variable. In that case, you can output a clear error message if no variable is set. There's no way to discover that this wrong until you make a request to serum-api

aspin (changes requested):
need to delete const AuthHeader

bxserum/connections/http.go
```diff
@@ -28,11 +30,13 @@ func (h HTTPError) Error() string {
 // HTTP response for GET request
 func HTTPGet[T protoreflect.ProtoMessage](url string, val T) error {
```
aspin: is this function used anywhere? If not can probably delete it

bxserum/provider/common.go
```diff
@@ -30,9 +33,10 @@ type RPCOpts struct {
 	Endpoint   string
 	Timeout    time.Duration
 	PrivateKey *solana.PrivateKey
+	AuthHeader string
 }
 
-func DefaultRPCOpts(endpoint string) RPCOpts {
+func DefaultRPCOpts(endpoint string, authHeader string) RPCOpts {
```
aspin: better to load `authHeader` from the environment variable in this function, to save lines throughout the rest of the changeset

examples/httpclient/main.go
aspin: change these back to `NewHTTPClient`

aspin (commented):
small changes requested. looks in good shape

aspin (approved):
lgtm once you fix up builds

PR #37, opened 2022-08-09
Adding openOrdersAddress to the GetOpenOrders methods

tinystarinagalaxy:


bxserum/provider/ws.go
```diff
@@ -80,7 +80,7 @@ func (w *WSClient) GetTickers(ctx context.Context, market string) (*pb.GetTicker
 }
 
 // GetOpenOrders returns all opened orders by owner address and market
-func (w *WSClient) GetOpenOrders(ctx context.Context, market string, owner string) (*pb.GetOpenOrdersResponse, error) {
+func (w *WSClient) GetOpenOrders(ctx context.Context, market string, owner string, openOrdersAddress string) (*pb.GetOpenOrdersResponse, error) {
```
aspin, line 83: need change on line 85 too

tinystarinagalaxy: done, do we have linter for these issues?

aspin: Do you know of any linter that would be helpful in this? I imagine we'd have to write a custom rule for this type of thing and I'm not really sure of how we could catch this automatically.

aspin (approved):

bxserum/provider/ws.go
tinystarinagalaxy: i have to google around. there are linters that we can set up with our github for sure. that catches all of these.

aspin: i suppose we should also consider not separating our arguments the way we are right now (`ctx context.Context, market string, owner string)`, and instead just pass in the protobuf object. If we did that then HTTP is the only interface that needs to be updated.

PR #40, opened 2022-08-10
Trad 28/add ci integration tests

DimaCantemir:
Trad 28/add ci integration tests
Setup CircleCI
Refactor to run SDK examples as Integration tests


.circleci/config.yml
```diff
@@ -0,0 +1,163 @@
...
+      - restore_cache:
+          keys:
+            - go-mod-{{checksum "go.sum"}}
+      - run:
+          name: Update/install packages
+          command: |
+            sudo apt update
+            sudo apt install awscli
+      - run:
+          name: Configure environment
```
aspin: think you can delete this step, + `GOPRIVATE` on line 12

.circleci/config.yml
```diff
@@ -0,0 +1,163 @@
...
+      - run:
+          name: Export build details
+          command: |
+            echo 'export COMMIT_HASH=$(git rev-parse HEAD)' >> $BASH_ENV
+            echo 'export COMMIT_AUTHOR="$(git --no-pager log -1 --pretty=format:'%an')"' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Unit test
+          command: make unit
+      - slack/notify:
```
aspin: Looks like this notifies Slack even on branches, which will probably be fairly noisy. IMO I think this is not needed for unit tests since you get pretty immediate feedback that will warn you against merging your PR branch in

.circleci/config.yml
```diff
@@ -0,0 +1,163 @@
...
+  version: 2
+  test-branch:
+    jobs:
+      - init:
+          context: circleci
+      - unit:
+          context: slack
+          requires:
+            - init
+      - integration:
```
aspin, line 114: add filter for develop only?

examples/grpcclient/main.go
```diff
@@ -52,125 +55,162 @@ func main() {
...
+	stream, err := g.GetOrderbookStream(ctx, []string{"SOL/USDC", "xxx"}, 3)
 	if err != nil {
-		log.Errorf("error with GetOrderbook stream request for SOL/USDC: %v", err)
-		return
+		log.Errorf("connection could not be established. error: %v", err)
+		return true
+	}
+
+	_, err = stream()
+	if err != nil {
```
aspin, line 180: fail if `err == nil`?

examples/grpcclient/main.go
```diff
@@ -202,25 +244,31 @@ const (
 	orderAmount = float64(0.1)
 )
 
-func orderLifecycleTest(g *provider.GRPCClient, ownerAddr string, ooAddr string) {
-	fmt.Println("\nstarting order lifecycle test")
+func orderLifecycleTest(g *provider.GRPCClient, ownerAddr string, ooAddr string) bool {
+	log.Info("\nstarting order lifecycle test")
```
aspin: remove leading `\n`?

aspin: couple more of these throughout and other `fmt.Println` or `fmt.Printf` lines

aspin (commented):
in general :+1:. Could do with a bit more cleanup of `fmt.Printf` => `log.Infof`, and perhaps getting rid of various `\n`s throughout. Also suggest to notify slack only on integration tests and only on develop

aspin (approved):

PR #45, opened 2022-09-07
TRAD-183 Go SDK update from recent block hash

yobloxroute:


bxserum/provider/grpc.go
```diff
@@ -63,10 +72,72 @@ func NewGRPCClientWithOpts(opts RPCOpts) (*GRPCClient, error) {
...
-	return &GRPCClient{
-		apiClient:  pb.NewApiClient(conn),
-		privateKey: opts.PrivateKey,
-	}, nil
+	client := &GRPCClient{
+		apiClient:                     pb.NewApiClient(conn),
+		privateKey:                    opts.PrivateKey,
+		recentBlockHashExpiryDuration: opts.CacheBlockHash,
+	}
+	go func() {
```
aspin: this code also needs to be in `provider/ws.go` –– probably would be a good idea to create a struct encapsulating this logic?

bxserum/provider/grpc.go
```diff
@@ -7,15 +7,24 @@ import (
...
 type GRPCClient struct {
 	pb.UnimplementedApiServer
 
+	mutex *sync.RWMutex
+
 	apiClient  pb.ApiClient
 	privateKey *solana.PrivateKey
+
+	recentBlockHash               *pb.GetRecentBlockHashResponse
+	recentBlockHashTime           *time.Time
```
aspin, line 26: any reason this needs to be a pointer? Seems to result in some unnecessary dereferencing

bxserum/provider/grpc.go
```diff
@@ -7,15 +7,24 @@ import (
...
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"sync"
+	"time"
 )
 
 type GRPCClient struct {
 	pb.UnimplementedApiServer
 
+	mutex *sync.RWMutex
```
aspin: any reason this needs to be a pointer?

bxserum/provider/grpc.go
aspin: It'd also be nice to have a boolean flag in RPCOpts for if the user wants to enable this. Can default to `true`

bxserum/provider/grpc.go
```diff
@@ -63,10 +72,72 @@ func NewGRPCClientWithOpts(opts RPCOpts) (*GRPCClient, error) {
...
+	response := g.getCachedBlockHash()
+	if response != nil {
+		return response, nil
+	}
+
+	g.mutex.Lock()
+	defer g.mutex.Unlock()
+
+	now := time.Now()
+	if g.recentBlockHash != nil && g.recentBlockHashTime.Before(now.Add(g.recentBlockHashExpiryDuration)) {
```
aspin: think the `.Before` call is backward –– should be `g.recentBlockHashTime.Add(g.recentblockHashExpiryDuration).Before(now)`?

aspin: `g.recentBlockHashTime` is already always before `now`

aspin (changes requested):

bxserum/provider/grpc.go
yobloxroute: just to be sure that we're not returning empty values, but yes, empty time is not valid :)

bxserum/provider/grpc.go
yobloxroute: 1. will think how to properly generalise it
2. basically we can use duration 0 for that, but flag will work too

bxserum/provider/grpc.go
yobloxroute: I mean time, for hash, which contains mutex it's necessary

PR #47, opened 2022-09-12
TRAD-209 add flag for filtering duplicates

aspin:


benchmark/internal/arrival/solanaws.go
```diff
@@ -165,22 +186,38 @@ func (s solanaOrderbookStream) Process(updates []arrival.StreamUpdate[solanaRawU
...
+		pu := ProcessedUpdate[SolanaUpdate]{
 			Timestamp: update.Timestamp,
 			Slot:      slot,
 			Data:      su,
-		})
+		}
+
+		redundant := su.IsRedundant()
+		if redundant {
+			duplicates[slot] = append(results[slot], pu)
```
kdonthi: Shouldn't this be `append(duplicates[slot])`...?
```suggestion
			duplicates[slot] = append(duplicates[slot], pu)
```

benchmark/serumapi/output.go
```diff
@@ -195,6 +217,11 @@ func PrintSummary(runtime time.Duration, serumEndpoint string, solanaEndpoint st
 			serumTs := matchedTs[0]
 			solanaTs := matchedTs[1]
 
+			// sometimes only 1 of asks and bids are matched
```
kdonthi, line 220: I am not sure I understand when this case would happen

kdonthi (approved):

benchmark/internal/arrival/solanaws.go
aspin: yep, mistake from me

benchmark/serumapi/output.go
aspin: saw some instances where there was only an update on the orderbook for bid or ask for a particular slot, but not both. for example, I think if someone cancels their bid that would cause this kind of case

PR #49, opened 2022-09-15
Updating SubmitTradeSwap methods to work with multiple transactions

tinystarinagalaxy:


aspin (approved):
change generally look fine, but want to see if we're sure it's necessary to split transactions

PR #50, opened 2022-09-20
TRAD-183 Go SDK update with recent block hash api

yobloxroute:


provider/ws.go
```diff
@@ -12,9 +12,10 @@ import (
 type WSClient struct {
 	pb.UnimplementedApiServer
 
-	addr       string
-	conn       *connections2.WS
-	privateKey *solana.PrivateKey
+	addr                 string
+	conn                 *connections2.WS
```
aspin: can you do a bit of refactoring and delete this alias?

provider/recent_hash_store.go
```diff
@@ -0,0 +1,98 @@
...
+		select {
+		case hash := <-ch:
+			s.updateBlockHash(hash)
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (s *recentBlockHashStore) updateBlockHash(hash *pb.GetRecentBlockHashResponse) {
```
aspin: can simplify all the functions in this struct's name -- it's already called `recentBlockHashStore`, so the `BlockHash` is implied. so methods can be `update()`, `get()` and `cached()`

provider/recent_hash_store.go
```diff
@@ -0,0 +1,98 @@
...
+		hashProvider:       hashProvider,
+		hashStreamProvider: streamProvider,
+		hash:               "",
+		hashTime:           time.Time{},
+		hashExpiryDuration: opts.CacheBlockHash,
+	}
+}
+
+func (s *recentBlockHashStore) run() {
+	ctx := context.Background()
```
aspin: should probably pass this context through the store when initializing it, as there's currently no way to cancel this loop

provider/recent_hash_store.go
```diff
@@ -0,0 +1,98 @@
...
+	response := s.getCachedBlockHash()
+	if response != nil {
+		return response, nil
+	}
+
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	now := time.Now()
+	if s.hash != "" && s.hashTime.Before(now.Add(s.hashExpiryDuration)) {
```
aspin: condition isn't necessary: `getCachedBlockHash` already checks this conditions

provider/recent_hash_store.go
aspin: I'm ok with requiring a `ctx` object for each provider type

provider/grpc.go
```diff
@@ -385,3 +402,13 @@ func (g *GRPCClient) GetOrderStatusStream(ctx context.Context, market, ownerAddr
...
 }
+
+// GetRecentBlockHashStream subscribes to a stream for getting recent block hash.
+func (g *GRPCClient) GetRecentBlockHashStream(ctx context.Context) (connections2.Streamer[*pb.GetRecentBlockHashResponse], error) {
+	stream, err := g.apiClient.GetRecentBlockHashStream(ctx, &pb.GetRecentBlockHashRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	return connections2.GRPCStream[pb.GetRecentBlockHashResponse](stream, ""), nil
```
aspin: can you also clean up the aliases in this file?

provider/common.go
```diff
@@ -37,10 +37,11 @@ type PostOrderOpts struct {
...
 type RPCOpts struct {
-	Endpoint   string
-	Timeout    time.Duration
-	PrivateKey *solana.PrivateKey
-	AuthHeader string
+	Endpoint       string
+	Timeout        time.Duration
+	PrivateKey     *solana.PrivateKey
+	AuthHeader     string
+	CacheBlockHash time.Duration
```
aspin: should also have a parameter to enable this. by default, the `recentBlockHashStore` should not run

aspin (changes requested):

Uri-Margalit-bloXroute (commented):
as a first timer seen those call methods it seems strange that the function returns `true` if it fails and `false` if succeeded .

The `failed = failed || function()` can be changed to `failed = failed && function()` and the functions can return `true` if succeeded

provider/ws.go
yobloxroute: you mean connections2? yeah, it looks strange will do

yobloxroute: done

provider/recent_hash_store.go
yobloxroute: done

provider/recent_hash_store.go
yobloxroute: yeah, in general it's worth even to pass context when creating client, that's not a part of the api

provider/recent_hash_store.go
yobloxroute: done

provider/grpc.go
yobloxroute: done

provider/common.go
yobloxroute: agree, changed

yobloxroute:
> as a first timer seen those call methods it seems strange that the function returns `true` if it fails and `false` if succeeded .
> 
> The `failed = failed || function()` can be changed to `failed = failed && function()` and the functions can return `true` if succeeded

believe it was done to accumulate error, but yeah, looks strange

aspin (approved):

PR #53, opened 2022-09-28
TRAD-230 Convert pools endpoint from POST to GET

yobloxroute:


DimaCantemir (approved):

examples/httpclient/main.go
```diff
@@ -179,6 +180,21 @@ func callTradesHTTP() bool {
 	return false
 }
 
+func callPoolsHTTP() bool {
+	h := provider.NewHTTPTestnet()
+
+	pools, err := h.GetPools([]string{"Radium"})
```
aspin, line 186: @yobloxroute does this example work? Doesn't pools need to be `P_RAYDIUM`?

DimaCantemir: yeah, it's going to fail

aspin: I think what actually ends up happening here is that it gets ignored. And at the moment (until my PR gets merged in), Pools endpoint doesn't actually check the requested project list

yobloxroute: Will go through endpoints tests, thank you for pointing this out!

yobloxroute: Yep, exactly, now I see it's fixed

PR #57, opened 2022-10-06
Implementing the RouteTradeSwap

tinystarinagalaxy:


DimaCantemir (approved):
lgtm

provider/ws.go
```diff
@@ -197,6 +197,16 @@ func (w *WSClient) PostTradeSwap(ctx context.Context, owner, inToken, outToken s
 	return &response, nil
 }
 
+// PostRouteTradeSwap returns a partially signed transaction for submitting a swap request
```
kdonthi: "Partially signed transaction(s)"

provider/grpc.go
```diff
@@ -200,14 +205,14 @@ func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, owner, inToken, outTok
...
-		return []string{}, err
+		return nil, err
 	}
 
 	var signatures []string
 	for _, tx := range resp.Transactions {
 		signature, err := g.signAndSubmit(ctx, tx, skipPreFlight)
 		if err != nil {
-			return signatures, err
+			return nil, err
```
kdonthi: It would be good to keep the signatures here so we can see what swaps have been completed

kdonthi (approved):

provider/grpc.go
aspin: +1. maybe good to wrap the error too, to include info about how many transactions were submitted?

aspin (approved):

provider/grpc.go
tinystarinagalaxy: we are returning signatures now.

provider/ws.go
tinystarinagalaxy: done

PR #58, opened 2022-10-08
TRAD-259 update endpoints

aspin:


provider/http.go
```diff
@@ -220,14 +220,14 @@ func (h *HTTPClient) signAndSubmit(tx string, skipPreFlight bool) (string, error
 
 // PostTradeSwap PostOrder returns a partially signed transaction for submitting a swap request
 func (h *HTTPClient) PostTradeSwap(owner, inToken, outToken string, inAmount, slippage float64, project pb.Project) (*pb.TradeSwapResponse, error) {
-	url := fmt.Sprintf("%s/api/v1/amm/trade-swap", h.baseURL)
+	url := fmt.Sprintf("%s/api/v1/trade/swap", h.baseURL)
```
kdonthi, line 223: Shouldn't quote and swap share more of the same base url?

kdonthi (approved):
Not sure about the base url change

provider/http.go
aspin: no, the existing separation in our API is `market` vs `trade`. `amm/` doesn't make sense in this context. The `quote` operation is meant to be part of the `market` API (e.g. reading data about the market), and `swap` is a `trade` operation (e.g. user want to make a trade)

PR #59, opened 2022-10-13
TRAD-32 Enable SSL on GRPC

aspin:
do a little bit of restructuring of the examples to support this

### DIFF
--- a/examples/httpclient/main.go
+++ b/examples/httpclient/main.go
@@ -16,14 +16,14 @@ import (
 )
 
 func httpClient() *provider.HTTPClient {
-	env, err := config.LoadEnv()
+	cfg, err := config.Load()
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	var h *provider.HTTPClient
 
-	switch env {
+	switch cfg.Env {
 	case config.EnvTestnet:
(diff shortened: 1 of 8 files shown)


examples/httpclient/main.go
```diff
@@ -431,9 +466,7 @@ func callReplaceByClientOrderID(ownerAddr, payerAddr, ooAddr string) bool {
 	log.Info("starting replace by client order ID test")
 	fmt.Println()
 
-	client := &http.Client{Timeout: time.Second * 60}
-	rpcOpts := provider.DefaultRPCOpts(provider.TestnetHTTP)
-	h := provider.NewHTTPClientWithOpts(client, rpcOpts)
+	h := httpClientWithTimeout(time.Second * 60)
```
atulsriv, line 469: Why are some timeouts 30s vs 60s?

atulsriv (approved):

tinystarinagalaxy (approved):

examples/httpclient/main.go
aspin: Some endpoints were taking longer compared to others. I think we've sped things up enough lately that this should mostly be unnecessary, but I'll leave that for another PR.

PR #60, opened 2022-10-13
running all steps and return error for each step

tinystarinagalaxy:
Not too excited about this change. but opened the PR for everyone to see and have discussion about it. Some clients might find this way useful, some might not like it.

### DIFF
--- a/provider/grpc.go
+++ b/provider/grpc.go
@@ -195,7 +195,7 @@ func (g *GRPCClient) PostSubmit(ctx context.Context, txBase64 string, skipPreFli
 }
 
 // SubmitTradeSwap builds a TradeSwap transaction then signs it, and submits to the network.
-func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, ownerAddress, inToken, outToken string, inAmount, slippage float64, project pb.Project, skipPreFlight bool) ([]string, error) {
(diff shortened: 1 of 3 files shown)


provider/grpc.go
```diff
@@ -205,48 +205,36 @@ func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, ownerAddress, inToken,
...
 		signature, err := g.signAndSubmit(ctx, tx, skipPreFlight)
-		if err != nil {
-			if signature != "" {
-				signatures = append(signatures, signature)
-			}
-			return signatures, err
-		}
-
 		signatures = append(signatures, signature)
+		errors = append(errors, err)
```
aspin, line 215: Perhaps we can write a common function for submit transactions, with some options? Whatever the case we'll have to write some documentation on this.

```go

type SubmitStrategy int

const (
    SubmitStrategyAll SubmitStrategy = iota  // submit all and collect signatures
    SubmitStrategyDeferLast // always execute last transaction even if there's an error
    SubmitStrategyAbortOnError // stop on submitting if simulation error
    SubmitStrategyWaitConfirmation // wait for full confirmation of previous transaction before moving to next, abort if error
)
```

We could also extend the submission endpoint to accept multiple transactions?

tinystarinagalaxy: yeah, this is a good idea.

PR #62, opened 2022-10-24
Adding Memo Hooks in sdk

tinystarinagalaxy:


transaction/memo.go
```diff
@@ -0,0 +1,144 @@
...
+	"fmt"
+	"github.com/gagliardetto/solana-go"
+	solanarpc "github.com/gagliardetto/solana-go/rpc"
+)
+
+const BxMemoMarkerMsg = "Powered by bloXroute Trader Api"
+
+var TraderAPIMemoProgram = solana.MustPublicKeyFromBase58("HQ2UUt18uJqKaQFJhgV9zaTdQxUZjNrsKFgoEDquBkcx")
+
+// CreateTraderAPIMemoInstruction generates a transaction instruction that places a memo in the transaction log
```
aspin, line 14: add more detail to the to the comment to indicate this is required for submitting transactions through Trader API

transaction/memo.go
```diff
@@ -0,0 +1,144 @@
...
+	instructions = append(instructions, memo)
+
+	txnBytes, err := buildPartialSignedTxn(blockhash, owner, instructions, privateKeys)
+	if err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(txnBytes), nil
+}
+
+func AddMemoToSerializedTxn(programIDIndex uint16, unsignedTxBase64, memoContent string,
```
aspin: add a note here that this primarily is for if the user wants to create the transaction on their own w/o using Trader API endpoints

aspin: haven't had much use for unit tests so far in this repo, but this file might be a good place for it?

transaction/memo.go
```diff
@@ -0,0 +1,144 @@
...
+		accounts := cmpInst.ResolveInstructionAccounts(&solanaTx.Message)
+		//decodedData, err := base58.Decode(cmpInst.Data)
+		instructions = append(instructions, &solana.GenericInstruction{
+			AccountValues: accounts,
+			ProgID:        progID,
+			DataBytes:     cmpInst.Data,
+		})
+	}
+	instructions = append(instructions, memo)
+	txnBytes, err := buildPartialSignedTxn(solanaTx.Message.RecentBlockhash, owner, instructions, privateKeys)
```
aspin: think there's no reason for partial signing here –– user should provide all the private keys that they used for the initial transaction and can fully complete signing

aspin (changes requested):
generally looks good! 👍

transaction/memo.go
tinystarinagalaxy: done

transaction/memo.go
tinystarinagalaxy: done

tinystarinagalaxy: yeah, i will add them by EOD. im testing the main example programs now

transaction/memo.go
tinystarinagalaxy: hmm, we have to sign the transaction and send back the base64 signed version though. 

Renamed to function to sign, and removed the -1 check. User needs to pass his privatekeys, so he gets a signed and ready base64 txn

transaction/memo.go
```diff
@@ -0,0 +1,134 @@
...
+	signedTxBytes, err := solanarpc.DataBytesOrJSONFromBase64(txBase64)
+	if err != nil {
+		return "", err
+	}
+	unsignedTx := solanarpc.TransactionWithMeta{Transaction: signedTxBytes}
+	solanaTx, err := unsignedTx.GetTransaction()
+	if err != nil {
+		return "", err
+	}
+	progID, err := solanaTx.Message.ResolveProgramIDIndex(programIDIndex)
```
aspin: A bit confused about what's happening here. What program index are we trying use here? Looking at like 70 below it looks like we're assigning the same program ID to every instruction, but isn't that incorrect in the case that the user has a transaction that interacts with multiple programs?

transaction/memo.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	memo := CreateTraderAPIMemoInstruction(memoContent)
+	var instructions []solana.Instruction
+
+	for _, cmpInst := range solanaTx.Message.Instructions {
+		accounts := cmpInst.ResolveInstructionAccounts(&solanaTx.Message)
+		instructions = append(instructions, &solana.GenericInstruction{
+			AccountValues: accounts,
+			ProgID:        progID,
```
aspin: this part confuses me –– looks like you should call `ResolveProgramIDIndex` for each instruction

transaction/memo.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	txBuilder.SetRecentBlockHash(recentBlockHash)
+	txBuilder.SetFeePayer(owner)
+
+	tx, err := txBuilder.Build()
+	if err != nil {
+		return nil, err
+	}
+	if privateKeys != nil {
```
aspin: lines below this point can be simplified with `tx.Sign()` function

transaction/memo.go
```diff
@@ -0,0 +1,134 @@
...
+
+	for _, cmpInst := range solanaTx.Message.Instructions {
+		accounts := cmpInst.ResolveInstructionAccounts(&solanaTx.Message)
+		instructions = append(instructions, &solana.GenericInstruction{
+			AccountValues: accounts,
+			ProgID:        progID,
+			DataBytes:     cmpInst.Data,
+		})
+	}
+	instructions = append(instructions, memo)
```
aspin: `return AddMemo(...)`?

aspin (changes requested):

transaction/memo.go
tinystarinagalaxy: fixed this. we are not assign the programID of each instruction properly

transaction/memo.go
tinystarinagalaxy: fixed this. we are not assign the programID of each instruction properly

transaction/memo.go
tinystarinagalaxy: done

transaction/memo.go
aspin: bump, the `solana-go` library can simplify all this code. only reason the rest of this exist elsewhere in the codebase is because we needed to partially sign it but we can use the built-in function here instead

tinystarinagalaxy: done

aspin: can now delete this `if` condition

aspin (approved):
one more small note

transaction/memo.go
tinystarinagalaxy: already cleaned this up

PR #63, opened 2022-10-26
Use submitBatch when submitting multiple transactions

tinystarinagalaxy:


provider/grpc.go
```diff
@@ -209,7 +209,11 @@ func (g *GRPCClient) PostSubmitBatch(ctx context.Context, request *pb.PostSubmit
 }
 
 // SubmitTradeSwap builds a TradeSwap transaction then signs it, and submits to the network.
-func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, ownerAddress, inToken, outToken string, inAmount, slippage float64, project pb.Project, skipPreFlight bool) ([]string, error) {
+func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, ownerAddress, inToken, outToken string, inAmount, slippage float64, project pb.Project, submitStrategy pb.SubmitStrategy, skipPreFlight bool) (*pb.PostSubmitBatchResponse, error) {
```
aspin: i think it'll be help us down the line to combine `submitStrategy` and `skipPreFlight` into an `SubmitOpts` struct w/ default values,.

provider/grpc.go
```diff
@@ -222,43 +226,52 @@ func (g *GRPCClient) SubmitTradeSwap(ctx context.Context, ownerAddress, inToken,
...
+		return nil, ErrPrivateKeyNotFound
+	}
+
+	resp, err := g.PostRouteTradeSwap(ctx, request)
 	if err != nil {
 		return nil, err
 	}
-
-	var signatures []string
+	batchRequest := pb.PostSubmitBatchRequest{}
```
aspin: this whole section can be refactor into a single function to be shared across the other submission endpoints, including the `g.privateKey == nil` check

aspin (changes requested):
- refactor submission into opts w/preflight + strategy
- refactor submitting multiple transactions, like `signAndSubmit`

provider/grpc.go
tinystarinagalaxy: good idea, done.

provider/grpc.go
tinystarinagalaxy: done

provider/http.go
```diff
@@ -219,6 +219,41 @@ func (h *HTTPClient) signAndSubmit(tx string, skipPreFlight bool) (string, error
...
+				return nil, err
+			}
+			oneRequest.Transaction = &pb.TransactionMessage{
+				Content:   signedTxBase64,
+				IsCleanup: txMsg.IsCleanup,
+			}
+		}
+
+		batchRequest.Entries = append(batchRequest.Entries, &oneRequest)
+	}
```
kdonthi: Can you refactor this? Doesn't look clean to have this big chunk for each provider

tinystarinagalaxy: we already follow the same patter for `signAndSubmit` also, im keeping it consistent.

kdonthi (changes requested):
I think creating one function for doing the submitted batch shared across the different providers would be clean.

provider/http.go
kdonthi: SignAndSubmit is like 15 lines with spaces, this is 30

Also, there isn't really anything to refactor across the different endpoints for signAndSubmit except for this:
```
        if w.privateKey == nil {
		return "", ErrPrivateKeyNotFound
	}

	txBase64, err := transaction.SignTxWithPrivateKey(tx, *w.privateKey)
	if err != nil {
		return "", err
	}
```

tinystarinagalaxy: making this `signAndSubmitBatch` more general than this, requires passing function pointer from each call, which makes calling this function not as smooth as it can be.  i can do it. but it will make the code less readable for not so much of a gain. 

we have chosen the right level of abstraction at this point. it's ok if each provider have simiiar functions. we can go ahead and merge most of their functions, using function pointers or extra interfaces, but it won't lead to any substantial benefit.

provider/grpc.go
```diff
@@ -165,6 +165,41 @@ func (g *GRPCClient) signAndSubmit(ctx context.Context, tx string, skipPreFlight
 	return response.Signature, nil
 }
 
+// signAndSubmitBatch signs the given transactions and submits them.
+func (g *GRPCClient) signAndSubmitBatch(ctx context.Context, transactions interface{}, opts SubmitOpts) (*pb.PostSubmitBatchResponse, error) {
```
aspin, line 169: does this have to be `interface{}` type? I think it's better to require `pb.TransactionMessage` type for this and update the other calls to force this

aspin (commented):
comment on signatures, looks good otherwise

provider/grpc.go
tinystarinagalaxy: yeah, it must be like that, we have `string transaction` sometimes. and sometimes we have `TransactionMessage transaction` and since we pass them as slice. we cannot even have `[]interface{}` in arg, first it must be `interface{}` then we iterate thru that and cast to the right type

aspin: isn't this function only for when you have multiple tranasctions, and all endpoints that return multiple transactions are `TransactionMessage`?

tinystarinagalaxy: discussed offline, that cancelAll uses `[]string` we can do this refactor in another ticket

aspin (approved):

kdonthi (approved):

PR #69, opened 2022-11-08
Make Transaction Return Type Consistent

atulsriv:


aspin (approved):

provider/grpc.go
```diff
@@ -63,7 +63,7 @@ func (bc blxrCredentials) RequireTransportSecurity() bool {
 
 // NewGRPCClientWithOpts connects to custom Trader API
 func NewGRPCClientWithOpts(opts RPCOpts) (*GRPCClient, error) {
-	authOption := grpc.WithPerRPCCredentials(blxrCredentials{authorization: opts.AuthHeader})
+	authOption := grpc.WithPerRPCCredentials(blxrCredentials{authorization: "ZDIxYzE0NmItZWYxNi00ZmFmLTg5YWUtMzYwMTk4YzUyZmM4OjEwOWE5MzEzZDc2Yjg3MzczYjdjZDdhNmZkZGE3ZDg5"})
```
aspin: remove this. probably should purge this commit too, though probably fine since we can just validate this header if need be. but be sure to be careful about commiting credentials

aspin (commented):
changes look good

aspin (commented):
need to fix up the examples too though i think

examples/grpcclient/main.go
```diff
@@ -93,6 +94,9 @@ func run() bool {
 	if !ok {
 		log.Infof("PAYER environment variable not set: will be set to owner address")
 		payerAddr = ownerAddr
+	} //ooAddr, ok := os.LookupEnv("OPEN_ORDERS")
```
aspin: delete?

PR #75, opened 2022-11-30
TRAD-381 Update solana-trader-api GO client

yobloxroute:


.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "solana-trader-proto"]
```
aspin: let's not do this quite yet

aspin (commented):
don't delete submodule yet, rest looks fine

.gitmodules
yobloxroute: returned back

aspin (approved):

aspin (approved):

PR #76, opened 2022-12-06
Integration with api

tinystarinagalaxy:


provider/http.go
```diff
@@ -14,15 +14,15 @@ import (
 type HTTPClient struct {
 	pb.UnimplementedApiServer
 
-	baseURL    string
-	httpClient *http.Client
-	requestID  utils.RequestID
-	privateKey *solana.PrivateKey
-	authHeader string
+	BaseURL    string
```
aspin: reasoning behind this? I don't see a need to expose any of these attributes outside of this package based on this PR

provider/http.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	HttpClient *http.Client
+	RequestID  utils.RequestID
+	PrivateKey *solana.PrivateKey
+	AuthHeader string
 }
 
 func (h *HTTPClient) GetAuthHeader() string {
```
aspin: nit: should be called `AuthHeader()`, or can be deleted entirely if we need to make the properties public

provider/http.go
aspin: ah, for integration tests? Does the constructor now provide good enough configurability?

tinystarinagalaxy: oh, these are left overs, will revert these.

provider/http.go
```diff
@@ -49,25 +49,36 @@ func NewHTTPLocal() *HTTPClient {
 	return NewHTTPClientWithOpts(nil, opts)
 }
 
+// NewHTTPCustomClient
+func NewHTTPCustomClient(authHeader, endpoint string) *HTTPClient {
```
tinystarinagalaxy, line 49: we use this in solana-trader api, don't need the props to be public

provider/http.go
tinystarinagalaxy: reverted

aspin (approved):

provider/http.go
tinystarinagalaxy: removed this func and made all props private

PR #81, opened 2023-01-03
TRAD-325 use ctx as part of HTTP client to control timeouts [breaking]

aspin:
breaks some old HTTP APIs, and now requires ctx to be part of each call
also includes the missing `includeFailed` option for swaps


tinystarinagalaxy (approved):

connections/http.go
```diff
@@ -44,13 +45,13 @@ func HTTPGetWithClient[T protoreflect.ProtoMessage](url string, client *http.Cli
...
 
-func HTTPPostWithClient[T protoreflect.ProtoMessage](url string, client *http.Client, body interface{}, val T, authHeader string) error {
+func HTTPPostWithClient[T protoreflect.ProtoMessage](ctx context.Context, url string, client *http.Client, body interface{}, val T, authHeader string) error {
 	b, err := json.Marshal(body)
 	if err != nil {
 		return err
 	}
 
-	req, err := http.NewRequest("POST", url, bytes.NewBuffer(b))
+	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewBuffer(b))
```
kdonthi, line 54: How does the context get sent with the request?

kdonthi: And we are checking the timeout of the context for each request on the API side?

kdonthi (approved):

connections/http.go
aspin: it's purely a client side timeout. So the SDK is the one that can set a timeout for the server to respond properly in, and if it doesn't then the SDK will return an error

PR #85, opened 2023-01-18
prevent accidental private key examples in CI

aspin:


.circleci/config.yml
```diff
@@ -58,13 +58,19 @@ jobs:
           at: /home/circleci/go
       - run:
           name: GRPC Examples (testnet)
-          command: API_ENV=testnet AUTH_HEADER=$AUTH_HEADER_TESTNET  make grpc-examples
+          command: |
+            unset PRIVATE_KEY
```
kdonthi, line 62: does this just prevent the private key in `.env` files?

kdonthi (approved):

atulsriv (approved):

.circleci/config.yml
aspin: There are no `.env` files in the integration tests. CircleCI context loads environment variables direct, kinda like what a `.env` file would do. But this specifically makes it so that even if `PRIVATE_KEY` is loaded from CircleCI context it will not be present after that line of code, so the private key will not be available to the examples and thus transactions can't be signed.

PR #89, opened 2023-02-06
Implementing the perp support in sdk

tinystarinagalaxy:


examples/config/env.go
```diff
@@ -10,6 +10,7 @@ type Example struct {
 	Env           Env
 	RunSlowStream bool
 	RunTrades     bool
+	RunDriftPosts bool
```
aspin: IMO we can combine this with `RunTrades`. But if you want them separate then let's update the naming a bit? `RunAmmOrderbookTrades` and `RunPerpTrades`?

provider/grpc.go
```diff
@@ -306,9 +328,8 @@ func (g *GRPCClient) SubmitOrder(ctx context.Context, owner, payer, market strin
 }
 
 // SubmitOrder builds a perp order, signs it, and submits to the network.
-func (g *GRPCClient) SubmitPerpOrder(ctx context.Context, owner, payer, accountAddress, slippage string, positionSide common.PerpPositionSide, typee common.PerpOrderType,
-	contract common.PerpContract, amount, price float64, project pb.Project, clientOrderID uint64, skipPreFlight bool) (string, error) {
-	order, err := g.PostPerpOrder(ctx, owner, payer, accountAddress, slippage, positionSide, typee, contract, amount, price, project, clientOrderID)
+func (g *GRPCClient) SubmitPerpOrder(ctx context.Context, request *pb.PostPerpOrderRequest, skipPreFlight bool) (string, error) {
```
aspin, line 326: I'm fine with this conversion and would actually like for us to do it in the long-term, but note that this going to be different from the rest of the functions

provider/grpc.go
```diff
@@ -139,6 +139,41 @@ func (g *GRPCClient) GetOpenOrders(ctx context.Context, market string, owner str
 	return g.apiClient.GetOpenOrders(ctx, &pb.GetOpenOrdersRequest{Market: market, Address: owner, OpenOrdersAddress: openOrdersAddress, Project: project})
 }
 
+// GetOpenPerpOrders returns all opened perp orders
+func (g *GRPCClient) GetOpenPerpOrders(ctx context.Context, owner string, accountAddress string, contracts []common.PerpContract, project pb.Project) (*pb.GetOpenPerpOrdersResponse, error) {
```
aspin: can you make this all consistent? Either make it all `request *pb.type` or enumerate the arguments in the function signature. Let's not try to avoid mixing it in new code

provider/ws.go
```diff
@@ -630,6 +647,43 @@ func (w *WSClient) SubmitReplaceOrder(ctx context.Context, orderID, owner, payer
 	return w.signAndSubmit(ctx, order.Transaction, opts.SkipPreFlight)
 }
 
+func (w *WSClient) SubmitDepositCollateral(ctx context.Context, req *pb.PostDepositCollateralRequest, skipPreFlight bool) (interface{}, interface{}) {
```
aspin: needs a better type signature than `interface{}, interface{}`

aspin (changes requested):
function signatures should be consistent. I'm fine with either all new methods passing in the entire protobuf type (and not having type out each argument) or the other way around, but not both

examples/config/env.go
tinystarinagalaxy: renamed to `RunPerpTrades`

provider/grpc.go
tinystarinagalaxy: made the change everywhere for perp functions

provider/grpc.go
tinystarinagalaxy: done

provider/ws.go
tinystarinagalaxy: fixed

aspin (approved):

PR #96, opened 2023-02-16
TRAD-556 Versioned transactions support

DimaCantemir:
TRAD-556 Versioned transactions support


atulsriv:
can we add a test to do a jupiter swap?

DimaCantemir:
> 

interestingly - no. I need this merged to complete the traderAPI PR that actually makes this work. Then I can merge and deploy traderAPI...and then I can add additional tests everywhere.

atulsriv (approved):

atulsriv:
> > 
> 
> interestingly - no. I need this merged to complete the traderAPI PR that actually makes this work. Then I can merge and deploy traderAPI...and then I can add additional tests everywhere.

Ok sounds good. I think we should have a test somewhere that does swaps (not only just the generation) whether that be in the api or in each sdk. It may be a bit tedious but probably useful...

kdonthi (approved):

aspin (approved):

PR #113, opened 2023-04-21
Adding Drift V2 endpoints Part1

tinystarinagalaxy:


aspin (approved):

provider/grpc.go
```diff
@@ -110,8 +110,8 @@ func (g *GRPCClient) GetRecentBlockHash(ctx context.Context) (*pb.GetRecentBlock
 }
 
 // GetOrderbook returns the requested market's orderbook (e.g. asks and bids). Set limit to 0 for all bids / asks.
-func (g *GRPCClient) GetOrderbook(ctx context.Context, market string, limit uint32, project pb.Project) (*pb.GetOrderbookResponse, error) {
-	return g.apiClient.GetOrderbook(ctx, &pb.GetOrderbookRequest{Market: market, Limit: limit, Project: project})
+func (g *GRPCClient) GetOrderbook(ctx context.Context, market string, limit uint32, metadata bool, project pb.Project) (*pb.GetOrderbookResponse, error) {
```
aspin: this field looks like it doesn't belong. Think you missed it as you switch from design 1 => 2?

provider/ws.go
```diff
@@ -73,9 +73,12 @@ func (w *WSClient) RecentBlockHash(ctx context.Context) (*pb.GetRecentBlockHashR
 }
 
 // GetOrderbook returns the requested market's orderbook (e.g. asks and bids). Set limit to 0 for all bids / asks.
-func (w *WSClient) GetOrderbook(ctx context.Context, market string, limit uint32, project pb.Project) (*pb.GetOrderbookResponse, error) {
+func (w *WSClient) GetOrderbook(ctx context.Context, market string, limit uint32, metadata bool, project pb.Project) (*pb.GetOrderbookResponse, error) {
```
aspin: here too

aspin (changes requested):
think a field has been left in that shouldn't be

provider/grpc.go
tinystarinagalaxy: removed

provider/ws.go
tinystarinagalaxy: removed

aspin (approved):

PR #114, opened 2023-04-25
add support for nil param rpc methods

atulsriv:
For format of json rpc subscribe calls as such: 

> {“jsonrpc”:“2.0", “id”:1, “method”: “eth_subscribe”, “params”: [“newPendingTransactions”]}
> {“jsonrpc”:“2.0”,“id”: 1, “method”: “eth_subscribe”, “params”: [“newHeads”]}

where we just need an array of ["newPendingTransaction"] and not ["newPendingTransactions": nil]

### DIFF
--- a/connections/ws.go
+++ b/connections/ws.go
@@ -264,7 +264,7 @@ func wsStream[T any](w *WS, ctx context.Context, streamName string, streamParams
 		StreamOpts: streamParams,
 	}
 
-	paramsB, err := json.Marshal(params)
+	paramsB, err := params.MarshalJSON()
 	if err != nil {
 		return nil, err
 	}
--- a/connections/ws.go
+++ b/connections/ws.go
@@ -225,10 +225,20 @@ func (w *WS) request(ctx context.Context, request jsonrpc2.Request, lockRequired
 }
 
 func WSStreamAny[T any](w *WS, ctx context.Context, streamName string, streamParams interface{}) (Streamer[T], error) {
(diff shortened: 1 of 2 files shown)


connections/ws.go
```diff
@@ -249,14 +259,26 @@ func WSStreamProto[T proto.Message](w *WS, ctx context.Context, streamName strin
...
-	}
-	paramsB, err := json.Marshal(params)
-	if err != nil {
-		return nil, err
+	var (
+		err     error
+		paramsB []byte
+	)
+	if streamParams == nil {
+		paramsB, err = json.Marshal([]string{streamName})
```
aspin: slightly better would be to edit `SubscribeParams`'s `MarshalJSON` function to check if `streamOpts` is nil, and to not marshal it there.

aspin (changes requested):
move code to a slightly better spot

connections/ws.go
```diff
@@ -259,24 +259,14 @@ func WSStreamProto[T proto.Message](w *WS, ctx context.Context, streamName strin
...
-		paramsB, err = json.Marshal(params)
-		if err != nil {
-			return nil, err
-		}
+	params := SubscribeParams{
+		StreamName: streamName,
+		StreamOpts: streamParams,
+	}
+
+	paramsB, err := params.MarshalJSON()
```
aspin: you still want to use `json.Marshal` here

aspin (approved):

PR #115, opened 2023-04-26
add retry logic to connection

atulsriv:


connections/ws.go
```diff
@@ -37,20 +39,22 @@ type WS struct {
 	// public to allow overriding of (un)subscribe method name
 	SubscribeMethodName   string
 	UnsubscribeMethodName string
+
+	endpoint   string
+	authHeader string
 }
 
 func NewWS(endpoint string, authHeader string) (*WS, error) {
```
aspin, line 47: probably should add boolean flag for if retries should be done

connections/ws.go
```diff
@@ -80,8 +97,22 @@ func (w *WS) readLoop() {
 
 		_, msg, err := w.conn.ReadMessage()
 		if err != nil {
-			_ = w.Close(err)
-			return
+			// reconnect the websocket connection if connection read message fails
+			for {
```
aspin, line 101: one edge case to consider that's a bit problematic: `writeLoop` currently looks like this:

```go
func (w *WS) writeLoop() {
	for {
		m := <-w.writeCh
		err := w.conn.WriteMessage(websocket.TextMessage, m)
		if err != nil {
			_ = w.Close(fmt.Errorf("error sending message: %w", err))
			return
		}
	}
}
```

If someone tries to write to the connection during reconnection, then that goroutine will exit and the connection struct will end up in a bad state.

Some options:
- have some managing goroutine that's responsible for spawning the `readLoop` and the `writeLoop`, (like go `w.readLoop(conn, ...)`)
- depending on the retry flag, keep the write goroutine around forever. though probably we actually should do a check for `<-ctx.Done` in both read/write loops

aspin (changes requested):

PR #124, opened 2023-05-31
TRAD-736 - Running Drift methods

tinystarinagalaxy:


.circleci/config.yml
```diff
@@ -60,17 +61,17 @@ jobs:
           name: GRPC Examples (testnet)
           command: |
             unset PRIVATE_KEY
-            API_ENV=testnet AUTH_HEADER=$AUTH_HEADER_TESTNET  make grpc-examples
+            API_ENV=testnet AUTH_HEADER=$AUTH_HEADER_TESTNET RUN_TRADES=$RUN_TRADES RUN_SLOW_STREAM=$RUN_SLOW_STREAM RUN_PERP_TRADES=$RUN_PERP_TRADES make grpc-examples
```
aspin: I think these are all unnecessary, since they're already exported to the environment. just adding to config.yml should be good

aspin (approved):
think you can take some changes out of the config.yml

.circleci/config.yml
tinystarinagalaxy: reverted

PR #126, opened 2023-06-02
TRAD-707: Migrating Drift endpoints to V2 part 1

tinystarinagalaxy:


provider/grpc.go
```diff
@@ -316,6 +316,21 @@ func (g *GRPCClient) PostOrder(ctx context.Context, owner, payer, market string,
 	})
 }
 
+// GetDriftPerpPositions returns all perp positions on Drift platform
+func (g *GRPCClient) GetDriftPerpPositions(ctx context.Context, request *pb.GetDriftPerpPositionsRequest) (*pb.GetDriftPerpPositionsResponse, error) {
```
aspin, line 320: here's something I'm thinking for V2 as well: let's try to simplify our interfaces, so we have to write less code. I'm not sure if we can get rid of these details for WS/HTTP in Go like we can in Python, but we can do something like this:

```go
type GRPCClient2 struct {
	pb.ApiClient
}

func NewGRPCClient2() (pb.ApiClient, error) {
	conn, err := grpc.Dial(...)
	if err != nil {
		return nil, err
	}
	return GRPCClient2{
		ApiClient: pb.NewApiClient(conn),
	}, nil
}

func main() {
	f, err := NewGRPCClient2()
	if err != nil {
		panic(err)
	}

	f.GetPerpPositions() // autogenerated
}
```

so we don't have write these all out. We can then even consider refactoring the example code to all use `pb.ApiClient` instead of having concrete implementations for GRPC/HTTP/WSClient

aspin (approved):

PR #137, opened 2023-07-26
TRAD804 update to latest proto 

atulsriv:


kdonthi (approved):

provider/common.go
```diff
@@ -151,3 +152,32 @@ func createBatchRequestEntry(opts SubmitOpts, txBase64 string, privateKey solana
 	}
 	return &oneRequest, nil
 }
+
+func convertProtoSideToString(side pb.Side) string {
+	switch side {
+	case pb.Side_S_UNKNOWN:
```
DimaCantemir: you are still using protobuf enums. The point of this was to phase PB enums out from everywhere not only in the communication layer.

provider/common.go
```diff
@@ -151,3 +152,32 @@ func createBatchRequestEntry(opts SubmitOpts, txBase64 string, privateKey solana
...
+		return "validation-error"
+	case pb.Side_S_BID:
+		return "bid"
+	case pb.Side_S_ASK:
+		return "ask"
+	default:
+		return "validation-error"
```
DimaCantemir: by validation error I meant trader api was supposed to return a validation error from the endpoint if Side's value was different then the 2 supported values

provider/common.go
```diff
@@ -151,3 +152,32 @@ func createBatchRequestEntry(opts SubmitOpts, txBase64 string, privateKey solana
...
+	}
+}
+
+func convertOrderTypeToString(orderType common.OrderType) string {
```
DimaCantemir: same as above, you convert from enum to strings but we are trying to get rid of the enums from our code. just use strings, define them as constants

provider/grpc.go
```diff
@@ -1083,12 +1083,13 @@ func (g *GRPCClient) GetMarketsV2(ctx context.Context) (*pb.GetMarketsResponseV2
 }
 
 // PostOrderV2 returns a partially signed transaction for placing a Serum market order. Typically, you want to use SubmitOrder instead of this.
-func (g *GRPCClient) PostOrderV2(ctx context.Context, owner, payer, market string, side pb.Side, amount, price float64, opts PostOrderOpts) (*pb.PostOrderResponse, error) {
+func (g *GRPCClient) PostOrderV2(ctx context.Context, owner, payer, market string, side pb.Side, orderType common.OrderType, amount, price float64, opts PostOrderOpts) (*pb.PostOrderResponse, error) {
```
DimaCantemir: change "side" and "type" here to be string. And do this everywhere

DimaCantemir (changes requested):

DimaCantemir (approved):

PR #139, opened 2023-08-02
Add Kline Method

kdonthi:
* Added methods and examples for `GetKline` for the three providers

### DIFF
--- a/provider/http.go
+++ b/provider/http.go
@@ -1392,7 +1392,7 @@ func (h *HTTPClient) GetTickersV2(ctx context.Context, market string) (*pb.GetTi
 
 // GetKline returns the requested market's candles over a certain time period and resolution
 func (h *HTTPClient) GetKline(ctx context.Context, market string, from time.Time, to time.Time, resolution time.Duration) (*pb.GetKlineResponse, error) {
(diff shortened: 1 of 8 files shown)


provider/http.go
```diff
@@ -1389,6 +1390,17 @@ func (h *HTTPClient) GetTickersV2(ctx context.Context, market string) (*pb.GetTi
 	return tickers, nil
 }
 
+// GetKline returns the requested market's candles over a certain time period and resolution
+func (h *HTTPClient) GetKline(ctx context.Context, market string, from time.Time, to time.Time, resolution time.Duration) (*pb.GetKlineResponse, error) {
+	url := fmt.Sprintf("%s/api/v1/market/kline/%s?from=%s&to=%s&resolution=%s", h.baseURL, market, from.UTC().Format(time.RFC3339), to.UTC().Format(time.RFC3339), resolution.String())
```
DimaCantemir, line 1395: this functionality is V2 only please make it V2 same think for trader-api - it needs to be implemented as V2 as we re going to remove support for v1 soon

DimaCantemir (changes requested):

PR #145, opened 2024-01-08
TRAD-863/priority-fee

etienneblxr:


atulsriv (changes requested):
A lot of functions in each provider and request need to be updated after the proto update is made. For example, 

```
func (h *HTTPClient) SubmitReplaceOrderV2(ctx context.Context, orderID, owner, payer, market string, side string, orderType string, amount, price float64, opts PostOrderOpts) (string, error) {
	order, err := h.PostReplaceOrderV2(ctx, orderID, owner, payer, market, side, orderType, amount, price, opts)
	if err != nil {
		return "", err
	}

	return h.SignAndSubmit(ctx, order.Transaction, opts.SkipPreFlight)
}
```

This function in HTTP client, needs the 2 parameters added and then added in the PostRepalceOrderV2 request after the proto update is made. 

Let me know if you have any further questions

PR #169, opened 2024-02-23
Add support for Raydium StepProject

DimaCantemir:
Add support for Raydium StepProject


atulsriv:
Do you think we should add an example using the quote response and parsing it? instead of all just hardcoding an ID?

atulsriv (approved):
See comment about adding an example decoding a quote response

DimaCantemir:
> Do you think we should add an example using the quote response and parsing it? instead of all just hardcoding an ID?

Ideally yes, but for now this should do it.  Or we's have to change the examples significantly and I didn't had time for this

PR #188, opened 2024-06-11
Setup for Go SDK to support retrieving new Raydium cpmm pools

808putnam:
Setup Notes:
 1. To test locally, you will need to pull the solanal-trader-proto branch TRAD-1031/support-raydium-new-cpmm-pool.
 2. Then use a replace command in the go.mod to point to the local proto version.

Details:
1. Adds in GetNewRaydiumPoolsStreamV2 for the GRPCClient and WSClient which adds additional parameter includeCPMM.
2. Refactors original GetNewRaydiumPoolsStream to call V2 version with includeCPMM set to false.

### DIFF
--- a/provider/grpc.go
+++ b/provider/grpc.go
@@ -36,6 +36,7 @@ func NewGRPCClient() (*GRPCClient, error) {
 // NewGRPCTestnet connects to Testnet Trader API
 func NewGRPCTestnet() (*GRPCClient, error) {
 	opts := DefaultRPCOpts(TestnetGRPC)
+	opts.UseTLS = true
 	return NewGRPCClientWithOpts(opts)
 }
 
--- a/provider/grpc.go
+++ b/provider/grpc.go
@@ -840,9 +840,18 @@ func (g *GRPCClient) GetSwapsStream(
 
 // GetNewRaydiumPoolsStream subscribes to a stream for getting recent swaps on projects & markets of interest.
 func (g *GRPCClient) GetNewRaydiumPoolsStream(
-	ctx context.Context,
+	ctx context.Context) (connections.Streamer[*pb.GetNewRaydiumPoolsResponse], error) {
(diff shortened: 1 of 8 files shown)


provider/grpc.go
```diff
@@ -840,9 +840,18 @@ func (g *GRPCClient) GetSwapsStream(
 
 // GetNewRaydiumPoolsStream subscribes to a stream for getting recent swaps on projects & markets of interest.
 func (g *GRPCClient) GetNewRaydiumPoolsStream(
-	ctx context.Context,
+	ctx context.Context) (connections.Streamer[*pb.GetNewRaydiumPoolsResponse], error) {
```
atulsriv: you should pass `includeCPMM` into here so that it is easier to test on the outside providers (i.e ws, grpc) instead of having to toggle it at the provider level

808putnam: But wouldn't that break existing clients as this is a public api?

I investigated go's support for optional parameters and parameters that have default values and saw that go does not support either of these. Hence, I went with the design pattern of adding in a new public api - GetNewRaydiumPoolsStreamV2 that adds in the ability to pass an includeCPMM parameter.

Note that the python and ts api surface allowed me to add in the includeCPMM parameter without having to add in a new public API. The python and ts sdk api's take a GetNewRaydiumPoolsRequest parameter while the go sdk does not.

Is there a better design pattern for this? Hmm, I'm wondering if we go with the v2 design pattern, should I change the parameter to be GetNewRaydiumPoolsRequest to future-proof it.

DimaCantemir: We don't do function overloads to support different signatures for different combination of flags in GO SDK. so I agree with @atulsriv

808putnam: Took out function overload. Added includeCPMM parameter to GetNewRaydiumPoolsStream.

atulsriv (approved):

DimaCantemir (approved):

PR #195, opened 2024-09-09
Update signAndSubmitBatch

Lanikean:


provider/grpc.go
```diff
@@ -283,6 +283,31 @@ func (g *GRPCClient) signAndSubmitBatch(ctx context.Context, transactions []*pb.
...
+
+	if len(transactions) == 1 {
+		txBase64, err := transaction.SignTxWithPrivateKey(transactions[0].Content, *g.privateKey)
+		if err != nil {
+			return nil, err
+		}
+
+		res, err := g.PostSubmit(ctx, &pb.TransactionMessage{
+			Content:   txBase64,
+			IsCleanup: false,
```
tinystarinagalaxy, line 295: plz make sure the `isCleanup`, `frontRunningProtection` and `useStakedRPCs` come from the request

tinystarinagalaxy: you need to do it for http and ws as well

provider/ws.go
```diff
@@ -541,6 +542,28 @@ func (w *WSClient) SignAndSubmitBatch(ctx context.Context, transactions []*pb.Tr
...
 		return nil, ErrPrivateKeyNotFound
 	}
+
+	if len(transactions) == 1 {
+		txBase64, err := transaction.SignTxWithPrivateKey(transactions[0].Content, *w.privateKey)
+		if err != nil {
+			return nil, err
+		}
+
+		res, err := w.PostSubmit(ctx, txBase64, *opts.SkipPreFlight, false, false)
```
tinystarinagalaxy, line 552: this `false,false` need to come from the request

provider/http.go
```diff
@@ -461,6 +461,28 @@ func (h *HTTPClient) SignAndSubmitBatch(ctx context.Context, transactions []*pb.
...
 		return nil, ErrPrivateKeyNotFound
 	}
+
+	if len(transactions) == 1 {
+		txBase64, err := transaction.SignTxWithPrivateKey(transactions[0].Content, *h.privateKey)
+		if err != nil {
+			return nil, err
+		}
+
+		res, err := h.PostSubmit(ctx, txBase64, *opts.SkipPreFlight, false, false)
```
tinystarinagalaxy, line 471: this `false,false` need to come from the request

provider/grpc.go
DimaCantemir: In this case IsCleanup will be FALSE always...so it's fine

DimaCantemir (approved):
LGTM

PR #198, opened 2024-09-12
Rclmm

atulsriv:


.gitignore
```diff
@@ -1,3 +1,6 @@
 .env
 .idea/
 *.csv
+
```
atulsriv, line 4: did you run go mod vendor? you shouldn't need to ever vendor in this repo

Lanikean: i might've back when i was exploring go stuff

Lanikean (approved):

PR #215, opened 2025-01-05
hacks to test daniel's scenario

atulsriv:


Lanikean (commented):
@atulsriv should we merge this as is, or formalize the hacky bits?

PR #235, opened 2025-04-15
Use protojson when marshalling HTTP requests

garrett-partenza-us:
The current implementation of timestamps in the HTTP provider has a bug that causes the timestamp field to not be marshaled correctly in the HTTP POST request implementation. This corrects that issue.

### DIFF
--- a/connections/http.go
+++ b/connections/http.go
@@ -3,7 +3,6 @@ package connections
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


connections/http.go
```diff
@@ -50,12 +50,21 @@ func HTTPGetWithClient[T protoreflect.ProtoMessage](ctx context.Context, url str
 }
 
 func HTTPPostWithClient[T protoreflect.ProtoMessage](ctx context.Context, url string, client *http.Client, body interface{}, val T, authHeader string) error {
-	b, err := json.Marshal(body)
+	protoMsg := body.(proto.Message)
```
DimaCantemir, line 53: why did you change the marshaler ? I vaguely remember we had issue with marshaling before. Don't remember details anymore. But this kinda change needs to be carefully tested before pushing to prod

DimaCantemir (approved):

PR #242, opened 2025-06-05
warn on TLS and do not default

garrett-partenza-us:
#### New Endpoint Naming Convention

* Two distinct endpoint names are now available for each DNS location:

  * For example: `MainnetNYHTTP` (non-TLS) and `MainnetNYHTTPSecure` (TLS).
* Previously, endpoints defaulted to TLS. This behavior has been corrected.

#### Provider Interface Updates

* Functions used to create providers have been updated to:

  * Check whether the endpoint uses TLS and issue a warning about potential performance implications.
  * Eliminate the default use of TLS unless explicitly specified by the client.

#### Connection Management

**gRPC:**

* Configured client keepalive parameters:

  * Keepalive ping interval: 15 seconds.
  * Keepalive timeout: 5 seconds.
  * `PermitWithoutStream` set to `true`, allowing keepalive pings even when there are no active streams.
* Note: `MaxConnectionIdle` is not applicable on the client side, but keepalives can achieve a similar effect.

**HTTP:**

* Set `IdleConnTimeout` to `0` to prevent idle connection closure.
* Enabled TCP keepalive using `net.Dialer` with:

  * Keepalive interval: 15 seconds.
  * Timeout: 5 seconds.
* Increased idle connection limits:

  * Max idle connections: 200.
  * Max idle connections per host: 20.

**WebSocket:**

* Updated the WebSocket dialer to:

  * Enable TCP keepalive with a 15-second interval.
  * Set a 5-second ping timeout.

#### Example Usage Updates

* Clients are no longer required to use TLS or the NY region by default when creating pump providers.
* Instead of calling fixed-region constructors like `New<ConnectionType>ClientPumpNY()`, clients should now use:

  * `New<ConnectionType>ClientPump(endpoint)` with configurable region and TLS options.

### DIFF
--- a/connections/ws.go
+++ b/connections/ws.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"sync"
 	"time"
@@ -79,7 +80,13 @@ func NewWS(endpoint string, authHeader string, disablePingLoop bool) (*WS, error
 }
 
 func connect(endpoint string, auth string) (*websocket.Conn, error) {
-	dialer := websocket.Dialer{HandshakeTimeout: handshakeTimeout}
+	dialer := websocket.Dialer{
+		HandshakeTimeout: handshakeTimeout,
+		NetDialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 15 * time.Second,
+		}).DialContext,
+	}
 	header := http.Header{}
 	header.Set("Authorization", auth)
 	header.Set("x-sdk", package_info.Name)
--- a/connections/ws.go
+++ b/connections/ws.go
@@ -83,7 +83,7 @@ func connect(endpoint string, auth string) (*websocket.Conn, error) {
 	dialer := websocket.Dialer{
 		HandshakeTimeout: handshakeTimeout,
 		NetDialContext: (&net.Dialer{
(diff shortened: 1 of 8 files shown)


atulsriv (approved):

connections/ws.go
```diff
@@ -79,7 +80,13 @@ func NewWS(endpoint string, authHeader string, disablePingLoop bool) (*WS, error
 }
 
 func connect(endpoint string, auth string) (*websocket.Conn, error) {
-	dialer := websocket.Dialer{HandshakeTimeout: handshakeTimeout}
+	dialer := websocket.Dialer{
+		HandshakeTimeout: handshakeTimeout,
+		NetDialContext: (&net.Dialer{
+			Timeout:   5 * time.Second,
+			KeepAlive: 15 * time.Second,
```
atulsriv, line 87: should we make this longer? temporal keeps it at 65 sec

atulsriv (approved):

connections/ws.go
garrett-partenza-us: Temporal keeps their server timeout at 65 seconds, not client.

garrett-partenza-us: Also the timeout simply refers to the timeout to setup the connection (five seconds is more than enough), and the keep alive refers to TCP-level keep alive pings to detect dead connections, not HTTP keep alive maximum duration.

PR #243, opened 2025-08-05
Examples fixes

DimaCantemir:
FasBestOption added to PostSubmitOpts
Recommended GRPC settings

### DIFF
--- a/examples/config/env.go
+++ b/examples/config/env.go
@@ -58,6 +58,7 @@ type Region string
 const (
 	NY Region = "ny"
 	UK Region = "uk"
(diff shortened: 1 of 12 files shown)


examples/config/env.go
```diff
@@ -58,6 +58,7 @@ type Region string
 const (
 	NY Region = "ny"
 	UK Region = "uk"
+	FR Region = "fr"
```
atulsriv, line 61: want to just add all the regions here? anything else breaks if we do this?

atulsriv (approved):

